PR #2, opened 2018-10-10
Handle autocomplete fields TRM 20774

dlrice:
Fixes sub-task issue "Handle autocomplete fields"  (TRM-20774). Expects as input data an array of objects with keys `label` and `term`. Upon user selection returns the text input value.

* Added key `term` to objects in existing `treeData` in [tree-data.js](https://github.com/ebi-uniprot/franklin-sites/blob/handle-autocomplete-fields-TRM-20774/src/app/common/tree-data.js). These are assumed to be unique.
* For component demo, the existing `treeData` in [tree-data.js](https://github.com/ebi-uniprot/franklin-sites/blob/handle-autocomplete-fields-TRM-20774/src/app/common/tree-data.js) is converted to a flattened array with the functions `getSquashedArrayOfPaths` and `getFlattenedPaths`.
* Moved tree-select styling to a common dropdown.scss.
* Created a separate component AutoCompleteItem to for each matching node to make using `scrollIntoView` easier.

### DIFF
--- a/src/styles/components/auto-complete.scss
+++ b/src/styles/components/auto-complete.scss
@@ -1,0 +1,7 @@
+@import './dropdown';
+
+.autcomplete-menu {
+    @extend .dropdown-menu;
+    top: 39px;
+    left: 0.5px;
+}
--- a/src/styles/components/auto-complete.scss
+++ b/src/styles/components/auto-complete.scss
@@ -3,5 +3,4 @@
 .autcomplete-menu {
     @extend .dropdown-menu;
     top: 39px;
-    left: 0.5px;
 }
--- a/__tests__/components/__snapshots__/autocomplete-item.spec.jsx.snap
+++ b/__tests__/components/__snapshots__/autocomplete-item.spec.jsx.snap
@@ -1,0 +1,17 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`AutocompleteItem component should render 1`] = `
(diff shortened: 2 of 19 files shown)


src/styles/components/auto-complete.scss
```diff
@@ -0,0 +1,6 @@
+@import './dropdown';
+
+.autcomplete-menu {
```
xwatkins: Typo

dlrice:
Addressed all issues:

Testing:
* Now passes all tests
* Create unit tests for
  * `getLastIndexOfSubstringIgnoreCase` function in `utils.js`.
  * Static methods of `autocomplete` and `autocomplete-item` components
* Created snapshots for `autocomplete` and `autocomplete-item` components

Styling:
* The caret is back by swapping span for button in `dropdown.scss`.
* Active nodes are now **bold**. Issue came from previously refactored `getFlattenedPaths` returning an array of arrays so to fix pulled out the first element.
* Height increased to what is expected via setting button `line-height: 1.6rem;` in `dropdown.scss`.
* Fixed auto-complete.scss filename typo. More generally, renamed auto-complete everywhere to autocomplete as this is actually a single world [actually a single world](https://en.wikipedia.org/wiki/Autocomplete)

dlrice:
Something I forgot to do was change the version. Do we want to bump from 0.0.3-dev.1 to 0.0.3-dev.2?

PR #3, opened 2018-10-17
Tree select autocomplete hybrid trm 20829 v2

dlrice:
Adds the option (via a prop) for the TreeSelect to places an autocomplete filter component above the standard node menu tree structure.

* Added the following props to Autocomplete:
  - **showDropwdownUpdated**: a callback to inform parent when the dropdown has disappeared (ie if no text input provided or if no items are match)
  - **clearOnSelect**: optional prop which when true will clear the text input when a selection has been made
  - **placeholder**: optional prop to place placeholder in text input
* Added the following props to TreeSelect:
  - **autocomplete**: optional prop which when true places an autocomplete filter at the top of the dropdown
  - **autocompletePlaceholder**: optional prop which is passed onto the Autocomplete component's placeholder (see above)
* Width of `.dropdown-menu` style is now 100%
* Any input within `.dropdown-menu` has `margin: 0`
* TreeSelect now uses node values (rather than labels) to find nodes (eg when deciding if a node is active/toggled)
* Added `itemLabel` and `pathLabel` to the output of `utils/getSquashedPath`'s array elements. Autocomplete displays the `pathLabel` while TreeSelect displays `itemLabel`.


src/utils.js
```diff
@@ -2,19 +2,34 @@ export function getLastIndexOfSubstringIgnoreCase(string, substring) {
...
       }
-    } else if (!id || thisNode.label === id) {
+    } else if (!value || thisNode.value === value) {
       flattened = [...flattened, nodePath];
     }
   });
   return flattened;
 };
+
+function getSquashedPath(items, sep = ' / ') {
```
xwatkins: This function should be tested

src/utils.js
```diff
@@ -2,19 +2,34 @@ export function getLastIndexOfSubstringIgnoreCase(string, substring) {
...
+  return {
+    value: items[items.length - 1].value,
+    pathLabel: items.map(item => item.label).join(sep),
+    itemLabel: items[items.length - 1].label,
+    items,
+
+  };
+}
+
+export function getSquashedArrayOfPaths(paths) {
```
xwatkins: This should be tested too

src/styles/components/dropdown.scss
```diff
@@ -5,13 +5,14 @@
...
   position: relative;
-  display: inline-block;
+  display: block;
   &>.button {
     margin: 0;
   }
   button {
     outline: 0;
   }
+  width: 15rem;
```
xwatkins: Should take width of parent

PR #6, opened 2018-12-07
Adding basic results table

xwatkins:


src/components/results-table.jsx
```diff
@@ -0,0 +1,35 @@
...
+        {columns.map(column => (
+          <div className="table-header" key={column.name}>
+            {column.label}
+          </div>
+        ))}
+      </div>
+    </div>
+    <div className="table-body">
+      {data.map((row, i) => (
+        <div className={i % 2 === 0 ? 'table-row' : 'table-row table-row-odd'}>
```
dlrice: These divs will also need unique keys.

dlrice: Wherever we are supposed to be using unique keys it could be easier to just the uuid package ie `key={v1()}` incase users have several tables on the same page from the same data. Or maybe could think about some combination of a uuid with data indexes.

PR #8, opened 2018-12-18
Trm 21054 row selection

dlrice:
Allow rows of the DataTable to be selected by creating a new column at the start of each row which has a checkbox. There is a slight issue in that the component demo only logs to console when a row has been selected and does not reflect the checked status in the checkbox.


xwatkins:
Looks good, wondering if it might be worth highlighting a row when it's been checked, maybe use $colour-light-blue with a $colour-white bottom border?

PR #10, opened 2018-12-21
Trm 21055 sort table

dlrice:
```
If a column is sortable (in the column object):
  display grey descend chevron in header next to column title
  entire column header is a clickable button, callback will tell parent the name of the clicked column

If a column is sorted by descend (in the column object):
  display black descend chevron in header next to column title
  entire column header is a clickable button, callback will tell parent the name of the clicked column

If a column is sorted by ascend (in the column object):
  display black ascend chevron in header next to column title
  entire column header is a clickable button, callback will tell parent the name of the clicked column
```

### DIFF
--- a/src/app/components.jsx
+++ b/src/app/components.jsx
@@ -103,11 +103,14 @@ const components = [
           label: 'Column 1',
           name: 'col1',
           render: row => <span>{row.fieldValue1.value}</span>,
+          sortable: true,
         },
         {
           label: 'Column 2',
           name: 'col2',
           render: row => <span>{row.fieldValue2.value}</span>,
+          sorted: 'ascend',
+          sortable: true,
         },
       ],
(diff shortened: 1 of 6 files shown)


xwatkins:
- I'm not sure whether it's still standard or not, but to `@import` an scss file without `.scss` it use to have to start with underscore (`_mixins.scss`)
- Could we have `cursor:pointer` on the table header button?
- Could we make the arrow slightly darker on `:hover` when not selected? You could use scss `darken()` https://robots.thoughtbot.com/controlling-color-with-sass-color-functions - I actually don't think there's enough contrast for $colour-seashell-grey so it might be best to start from $colour-black and have it lighten by 40% for unselected and lighten by 20% for hover?

Once this is done, feel free to merge the PR yourself ;-)

PR #12, opened 2019-02-12
Add "page intro" pattern

xwatkins:
Adding a pattern to display namespace information at the top of results pages. The name of the namespace works as a trigger to toggle the display of extra information, as well as a list of links (internal or external).

This branch is required before merging of `uniprot-website`

### DIFF
--- a/src/components/page-intro.jsx
+++ b/src/components/page-intro.jsx
@@ -7,12 +7,7 @@ import ChevronDown from '../svg/chevron-down.svg';
 import ChevronUp from '../svg/chevron-up.svg';
 
 class PageIntro extends Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      showContent: false,
-    };
-  }
+  state = { showContent: false };
 
   render() {
     const {
--- a/src/components/page-intro.jsx
+++ b/src/components/page-intro.jsx
@@ -36,7 +36,9 @@ results
           )}
         </h2>
 
-        <div className="intro-content" style={{ display: showContent ? 'block' : 'none' }}>
+        <div
+          className={showContent ? 'intro-content intro-content--display-content' : 'intro-content'}
+        >
           {children}
           <div className="intro-links">
             {links.map(link => (
--- a/src/components/page-intro.jsx
+++ b/src/components/page-intro.jsx
@@ -1,0 +1,71 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Link } from 'react-router-dom';
+import { formatLargeNumber } from '../utils';
+import '../styles/components/page-intro.scss';
+import ChevronDown from '../svg/chevron-down.svg';
+import ChevronUp from '../svg/chevron-up.svg';
+
+class PageIntro extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      showContent: false,
+    };
+  }
+
+  render() {
+    const {
+      title, resultsCount, children, links,
+    } = this.props;
+    const { showContent } = this.state;
+    return (
+      <div className="page-intro">
+        <h2>
+          <button
+            type="button"
+            onClick={() => this.setState({ showContent: !showContent })}
+            className="dropdown"
+          >
+            {showContent ? <ChevronUp /> : <ChevronDown />}
+            {' '}
+            {title}
+          </button>
+          {resultsCount > 0 && (
+          <small>
+            {' '}
+            {formatLargeNumber(resultsCount)}
+            {' '}
+results
+          </small>
+          )}
+        </h2>
+
+        <div className="intro-content" style={{ display: showContent ? 'block' : 'none' }}>
+          {children}
+          <div className="intro-links">
+            {links.map(link => (
+              <Link to={link.destination} key={link.title}>
+                {link.title}
+              </Link>
+            ))}
+          </div>
+        </div>
+      </div>
+    );
+  }
+}
+
+PageIntro.propTypes = {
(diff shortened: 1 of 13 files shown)


src/components/page-intro.jsx
```diff
@@ -0,0 +1,71 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Link } from 'react-router-dom';
+import { formatLargeNumber } from '../utils';
+import '../styles/components/page-intro.scss';
+import ChevronDown from '../svg/chevron-down.svg';
+import ChevronUp from '../svg/chevron-up.svg';
+
+class PageIntro extends Component {
+  constructor(props) {
```
its-mahdi: The `constructor` method here doesn't do anything besides creating the `state` object; Therefore, this can be removed and `state` can be initiated as a class property.

src/components/page-intro.jsx
```diff
@@ -0,0 +1,71 @@
...
+          <button
+            type="button"
+            onClick={() => this.setState({ showContent: !showContent })}
+            className="dropdown"
+          >
+            {showContent ? <ChevronUp /> : <ChevronDown />}
+            {' '}
+            {title}
+          </button>
+          {resultsCount > 0 && (
```
its-mahdi: I would suggest to put `resultsCount > 0` in a pair of parenthesis, so the conditional statement would be more obvious e.g. `(resultsCount > 0) && ...`

src/components/page-intro.jsx
```diff
@@ -0,0 +1,71 @@
...
+          <small>
+            {' '}
+            {formatLargeNumber(resultsCount)}
+            {' '}
+results
+          </small>
+          )}
+        </h2>
+
+        <div className="intro-content" style={{ display: showContent ? 'block' : 'none' }}>
```
its-mahdi: Maybe instead of inline-styling, we can have a BEM Modifier CSS class instead e.g. `intro-content--display-content`.

src/components/page-intro.jsx
```diff
@@ -0,0 +1,71 @@
...
+          {children}
+          <div className="intro-links">
+            {links.map(link => (
```
its-mahdi: I don't know the coding convention we follow here, but instead of `link` we can have `({ title, destination })`. If we follow this everywhere, the help on more complicated components will be more pronounced.

src/styles/components/page-intro.scss
```diff
@@ -0,0 +1,26 @@
+@import "../settings";
+@import "../colours";
+
+.page-intro {
+  .intro-content {
+    -webkit-column-count: 2; /* Chrome, Safari, Opera */
```
its-mahdi: It's better to leave the browser specific features to CSS post-processors. If we haven't got one set up, it might be worth setting it up now, otherwise these browser specific styles need to be added/removed and generally maintained with the rest of code.

src/styles/components/page-intro.scss
```diff
@@ -0,0 +1,26 @@
...
+    text-align: right;
+    a {
+      margin-right: $global-margin;
+    }
+  }
+  button {
+    cursor: pointer;
+    color: $colour-black;
+  }
+  button:focus {
```
its-mahdi: The `button:focuse` should be moved inside the `button` block -- above, and should be written as:

    button {
      &:focus {
        ...
      }
    }

src/components/page-intro.jsx
xwatkins: Prettier removes it with the current config. That's something we could discuss when we talk about the code styling decisions.

src/components/page-intro.jsx
xwatkins: Done

src/components/page-intro.jsx
xwatkins: Yup

src/styles/components/page-intro.scss
xwatkins: Woops

src/styles/components/page-intro.scss
xwatkins: I agree, but am not sure if our current SCSS/CSS loading processors are doing this. Could you make an GitHub issue plse?

src/components/page-intro.jsx
xwatkins: I think in general it's nicer to use destructuring, but in this case `title` is already destructured as another variable in the upper scope. I could rename it while destructuring but I think that makes things a bit more confusing...

src/components/page-intro.jsx
its-mahdi: It took me a second to realise what does 'Prettier' mean here! Let's chat about it later then!

src/components/page-intro.jsx
```diff
@@ -41,7 +36,9 @@ results
           )}
         </h2>
 
-        <div className="intro-content" style={{ display: showContent ? 'block' : 'none' }}>
+        <div
+          className={showContent ? 'intro-content intro-content--display-content' : 'intro-content'}
```
its-mahdi: Just a suggestion, up to you if you want to make it this way or not:
    className={`intro-content ${showContent && 'intro-content--display-content'}`}

src/components/page-intro.jsx
its-mahdi: Okay, fair enough.

src/styles/components/page-intro.scss
its-mahdi: Yes, I will make an issue for it. We currently don't have any proper CSS post-processing in place.

PR #14, opened 2019-02-26
Info list and Card components

xwatkins:
Adding 2 new patterns to the library. This is required by the `uniprotkb-entry` branch on uniprot-website


src/components/info-list.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import '../styles/components/info-list.scss';
+
+const InfoList = ({ infoData }) => (
+  <div className="info-list">
+    {infoData.map(
+      // Only draw if there is content
+      item => typeof item.content !== 'undefined' && (
```
its-mahdi: This line can be shorten to:

    item => item.content && (

PR #16, opened 2019-03-15
feat(card): Add `links` option.

vivekimsit:


xwatkins:
Looks good! Only one thing, maybe not so easy to notice on the mockups: the line under the title changes colour depending on the type of card. So default is grey, as it is now, but we need to be able to set it as an optional prop too.
Have also changed the background colour of the links to `seashell-grey` as it's more easily readable.

src/components/card.jsx
```diff
@@ -3,15 +3,61 @@ import PropTypes from 'prop-types';
 
 import '../styles/components/card.scss';
 
-const Card = ({ title, children }) => (
+const DEFAULT_BORDER_COLOR = '#e5e5e5';
```
xwatkins: It would be best if the default was taken from the css so it can be set as a SASS variable (I believe the `style` tag overwrites the default so you should be able to leave it empty if you want to use the css style?)

vivekimsit: Done.

PR #22, opened 2019-06-07
Add different button styles, and new icons

xwatkins:
Pretty simple atomic stuff. Adding documentation for different styles of buttons, as well as some tweaks to for icons to display nicely in buttons.


src/app/Typography.jsx
```diff
@@ -33,12 +33,12 @@ const Typography = () => (
 
     <section>
       <h1>Headers</h1>
-      <h1>H1. This is a title</h1>
-      <h2>H2. This is a title</h2>
-      <h3>H3. This is a title</h3>
-      <h4>H4. This is a title</h4>
-      <h5>H5. This is a title</h5>
-      <h6>H6. This is a title</h6>
+      <h1>H1. This is a title .48px</h1>
```
dlrice: Could be interpreted as being 0.48px.

src/styles/common/_buttons.scss
```diff
@@ -0,0 +1,43 @@
...
+    width: 0.9rem;
+    height: 0.9rem;
+    margin-top: -0.2rem;
+    margin-right: 0.5rem;
+    vertical-align: middle;
+    fill: $colour-white;
+  }
+}
+
+.button.secondary {
```
dlrice: Would `.button.secondary` and `.button.link-button` be more "BEM"-like if made as children of `.button`?

src/styles/components/in-page-nav.scss
```diff
@@ -17,7 +17,7 @@
 
   &--active a {
     color: $colour-dark-blue;
-    font-weight: 600;
+    font-weight: 700;
```
dlrice: Would font-weight be better as shared global variable?

PR #23, opened 2019-06-07
Trm 22173/data list

dlrice:
* Created `ScrollItemsLoader`:
	* detects:
		* when a user has scrolled to the bottom of scrollable container and calls the `loadMoreData`
		* if the scrollable container is not yet scrollable (ie not enough items have been loaded), request `loadMoreData`
	* the prop hasMoreItems informs `ScrollItemsLoader` when to stop requesting
	* designed to be flexible so can pass in your own:
		* scrollable container
		* list of items
* Updated `Card` component:
	* Title is optional
	* Can be selectable with checkbox
* Updated react version to use hooks
* Installed dev dep `@testing-library/react` to test hooks
* Added implicit-arrow-linebreak to eslint rules to prevent prettifier creating lines longer than 100 chars, thus failing the eslint


src/components/scroll-items-loader.jsx
```diff
@@ -0,0 +1,97 @@
...
+  }, [hasMoreItems]);
+
+  return cloneElement(
+    scrollContainer,
+    {
+      ref,
+      onScroll: checkLoadMoreItems,
+      className: `${scrollContainerClassName || ''} scroll-container`,
+    },
+    [items.length > 0 && items, (loading || items.length === 0) && loadingIndicator],
```
xwatkins: This is a bit hard to read, could `[items, loading  && loadingIndicator]` do the same thing?

src/app/ScrollItemsLoaderDemo.jsx
```diff
@@ -0,0 +1,66 @@
...
+        idKey={idKey}
+        onSelect={id => setSelected({ ...selected, [id]: !selected[id] })}
+        selected={selected}
+      />
+    </Fragment>
+  );
+};
+
+const ScrollItemsLoaderDemo = () => (
+  <DefaultPageLayout title="Franklin - Data Table" content=<ScrollItemsLoaderDemoContent /> />
```
xwatkins: I think `content=<ScrollItemsLoaderDemoContent />` should be wrapped in curly brackets? `content={<ScrollItemsLoaderDemoContent />}`
Not sure why the `jslint` target is not complaining

PR #25, opened 2019-06-25
Data loader hoc

xwatkins:
Make the data loader an HOC to be used by `DataList` and `DataTable`. Tests updated and coverage checked.


src/components/__tests__/data-table.spec.jsx
```diff
@@ -8,7 +8,7 @@ describe('DataTable', () => {
   const onSelect = jest.fn();
   const onHeaderClick = jest.fn();
   const selectable = true;
-  const selected = {};
+  const selected = { id0: true };
```
its-mahdi: I suppose `id0` is NOT a typo and the proper format for the index, right?

src/components/__tests__/data-table.spec.jsx
```diff
@@ -31,7 +31,7 @@ describe('DataTable', () => {
     },
   ];
   const data = fillArray(10, (element, index) => ({
-    [idKey]: `id-${index}`,
+    [idKey]: `id${index}`,
```
its-mahdi: Was there any specific reasons to remove the `-` from `id-${index}`?

src/components/data-loader.jsx
```diff
@@ -0,0 +1,88 @@
...
+  } = props;
+
+  const [loading, setLoading] = useState(false);
+  const [loadMoreItems, setLoadMoreItems] = useState(false);
+  const ref = useRef();
+  const loadingMessageId = v1();
+
+  useEffect(() => {
+    if (hasMoreData && isNotScrollable(ref)) {
+      // Note: @dan, commenting the following line
```
its-mahdi: There shouldn't be personal messages/comments in the code for other team members. If the functionality can be tested and is still correct after `setLoading(true)` is removed, then it should be removed. If it can't be indicated if this is a dead code or not, the comment can be written as: `// Is this function call still necessary?`.

src/components/data-table.jsx
```diff
@@ -210,33 +130,47 @@ DataTable.propTypes = {
    * An object which indicates which rows have been selected by the user.
    */
   selected: PropTypes.instanceOf(Object),
```
its-mahdi: Can the shape of `selected` be defined here or the problem is the dynamic keys?

src/components/data-list.jsx
```diff
@@ -1,57 +1,61 @@
...
+   * callback.
    */
-  hasMoreData: PropTypes.bool.isRequired,
+  onSelect: PropTypes.func,
   /**
-   * An array of JSX elements which will populate the scrollContainer.
+   * An object which indicates which rows have been selected by the user.
    */
-  items: PropTypes.arrayOf(PropTypes.element).isRequired,
+  selected: PropTypes.instanceOf(Object),
```
its-mahdi: Can the shape of `selected` be defined here or the problem is the dynamic keys?

src/components/data-table.jsx
```diff
@@ -210,33 +130,47 @@ DataTable.propTypes = {
...
 };
 
-export default DataTable;
+const DataTable = ({
+  data, columns, onSelect, selected, idKey, selectable, onHeaderClick,
+}) => (
+  <Fragment>
+    <table className="data-table__table">
+      <DataTableHead
+        {...{
```
its-mahdi: Should the object be spread like that?

src/components/data-table.jsx
```diff
@@ -210,33 +130,47 @@ DataTable.propTypes = {
...
+          selectable,
+          columns,
+          onHeaderClick,
+        }}
+      />
+      <DataTableBody
+        {...{
```
its-mahdi: Same as comment above.

src/components/info-list.jsx
```diff
@@ -2,17 +2,18 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import '../styles/components/info-list.scss';
-
+// NOTE: DELETE? is this used???
```
its-mahdi: Can this be tested and if it is not used remove from the code?

src/app/DataViews.jsx
```diff
@@ -97,35 +96,28 @@ const DataViewsContent = () => {
           <DataList
             {...{
```
its-mahdi: Is it necessary to spread the object like that? Couldn't it be written as:

    {
        data,
        idKey,
        onSelect,
        hasMoreData,

src/app/DataViews.jsx
```diff
@@ -97,35 +96,28 @@ const DataViewsContent = () => {
...
-              )),
+              selectable,
+              selected,
+              dataRenderer: content => <Fragment>{Object.values(content)}</Fragment>,
             }}
           />
         )}
         {view === TABLE && (
           <DataTable
             {...{
```
its-mahdi: Same as comment above.

src/components/data-list.jsx
```diff
@@ -1,57 +1,61 @@
...
-        onScroll={() => checkLoadMoreItems(loading, hasMoreData, ref, setLoadMoreItems)}
-        ref={ref}
+const DataList = ({
+  data, idKey, selectable, selected, onSelect, dataRenderer,
+}) => (
+  <Fragment>
+    {data.map(({ [idKey]: id, ...content }) => (
+      <Card
+        key={id}
+        {...{
```
its-mahdi: Should the object be spread like that?

src/components/__tests__/data-list.spec.jsx
```diff
@@ -4,35 +4,72 @@ import DataList from '../data-list';
 import { fillArray } from '../../utils';
 
 describe('DataList', () => {
-  const items = fillArray(10, (element, index) => <p key={index}>{index}</p>);
+  const data = fillArray(10, (element, index) => ({
+    id: `id-${index}`,
```
its-mahdi: What is the correct format for the id: With `-` or without?

src/components/__tests__/data-table.spec.jsx
xwatkins: correct, see below

src/components/__tests__/data-table.spec.jsx
xwatkins: Yes, because of the "selected" one above.

src/components/data-list.jsx
xwatkins: Dynamic keys but we can use `shapeOf` to check the values

src/components/info-list.jsx
xwatkins: That was a mistake, it is still used.

xwatkins:
Most comments addressed.

PR #29, opened 2019-07-19
Trm 21886 accordion

dlrice:
Created a single accordion element. Multiple can be grouped using the `.accordion-group` class.


src/styles/components/accordion.scss
```diff
@@ -0,0 +1,50 @@
+@import '../colours';
+
+.accordion {
+  width: 100%;
+  border: 1px solid $medium-gray;
```
xwatkins: Not sure where this colour is defined (probably Foundation) but we use the ones in `./colours.scss` - in this case it would be `$colour-light-grey;`

dlrice: Updated

PR #30, opened 2019-07-29
Add columns option for InfoList component

xwatkins:
Use css columns if a flag is specified as a prop and the screen is bigger than `small`

### DIFF
--- a/src/app/components.jsx
+++ b/src/app/components.jsx
@@ -182,6 +182,10 @@ const components = [
           title: 'Another item',
           content: <div>Some more content</div>,
         },
+        {
+          title: 'Yet another item',
+          content: loremIpsum({ count: 25, units: 'words' }),
+        },
       ],
     },
   },
--- a/src/components/info-list.jsx
+++ b/src/components/info-list.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 
 import '../styles/components/info-list.scss';
 
-const InfoList = ({ infoData, columns = false }) => (
+const InfoList = ({ infoData, columns }) => (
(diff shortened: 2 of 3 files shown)


src/app/components.jsx
```diff
@@ -182,6 +182,10 @@ const components = [
           title: 'Another item',
           content: <div>Some more content</div>,
         },
+        {
+          title: 'Yet another item',
+          content: loremIpsum({ count: 25, units: 'words' }),
```
dlrice: `content` is a string while the prop type is `PropTypes.element`.

src/components/info-list.jsx
```diff
@@ -3,8 +3,8 @@ import PropTypes from 'prop-types';
 
 import '../styles/components/info-list.scss';
 
-const InfoList = ({ infoData }) => (
-  <div className="info-list">
+const InfoList = ({ infoData, columns = false }) => (
```
dlrice: The default value for `columns` is specified in the function parameters and also in `defaultProps`. Would be good to stick to one place.

src/styles/components/info-list.scss
```diff
@@ -24,6 +28,23 @@
...
+      }
+    }
+  }
+}
+
+@include breakpoint(medium) {
+  .info-list--columns {
+    column-count: 2;
+    column-width: 50%;
+    .info-list__item {
```
dlrice: Is `item` a modifier of `columns` or just a different element?

xwatkins: `columns` is the modifier of the top level element, `item` doesn't change

PR #32, opened 2019-08-01
Trm 22502/bubble number

xwatkins:
Adding different ways to represent numbers


src/styles/components/doughnut-chart.scss
```diff
@@ -0,0 +1,85 @@
...
+      transform-origin: 0 50%;
+      left: $size;
+      width: $size;
+      height: $size * 2;
+      border-top-left-radius: 0;
+      border-bottom-left-radius: 0;
+    }
+  }
+
+  &__right-wrap {
```
dlrice: A lot of `&__left-wrap` and `&__right-wrap` are in common - I wonder if we could create base style and extend them?

src/components/long-number.jsx
```diff
@@ -0,0 +1,14 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { formatLargeNumber } from '../utils';
+
+const LongNumber = ({ number }) => <Fragment>{formatLargeNumber(number)}</Fragment>;
```
dlrice: As there is only one params into the component I am wondering if it would be better to have this a child? Ie
`<LongNumber>1000000</LongNumber>`?

PR #33, opened 2019-08-01
Trm 22702/filter search

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-22702

### DIFF
--- a/src/app/components.jsx
+++ b/src/app/components.jsx
@@ -5,6 +5,7 @@ import { loremIpsum } from 'lorem-ipsum';
 import { treeData, flattenedPaths } from './common/tree-data';
 import facetData from './common/facetData';
 import sequenceData from './common/sequence-data';
+import { getLipsumObjectArray } from './common/lipsum';
 import {
   Accordion,
   Autocomplete,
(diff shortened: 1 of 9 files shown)


src/styles/components/search-input.scss
```diff
@@ -0,0 +1,12 @@
+.search-input {
+  position: relative;
+  width: 100%;
+  display: inline-block;
+
+  &__icon {
```
xwatkins: It might be more flexible to define this as a class like `right-content` and it's better to use flexbox than absolute positioning for this.

src/components/search-input.jsx
```diff
@@ -0,0 +1,22 @@
...
+const SearchInput = props => (
+  <span className="search-input">
+    <input type="text" {...props} />
+    <SearchIcon className="search-input__icon" width={14} height={14} />
+  </span>
+);
+
+SearchInput.propTypes = {
+  // eslint-disable-next-line react/forbid-prop-types
+  props: PropTypes.object,
```
xwatkins: Is it not possible to define/restrict the content of these props more?

src/app/components.jsx
```diff
@@ -118,6 +119,15 @@ const components = [
       onSelect: () => {},
     },
   },
+  {
+    name: 'Search Input',
+    component: SearchInput,
+    function: 'blah',
```
xwatkins: Maybe something like:

Function: Filter data displayed on the screen (autocomplete, list of fields etc..)
Purpose: Provide the user with an indication that typing something in the input box will search something

PR #36, opened 2019-09-12
Use expandable list component and tweak styling

xwatkins:
- Add the expandable list component to facets so they are collapsed if more than 5 items.
- Tweak colours
- Add 'x' on hover when active to indicate clicking will remove the filter


src/app/common/facetData.js
```diff
@@ -37,6 +37,67 @@ const facetData = [
     label: 'Facet not shown',
     name: 'facet_not',
   },
+  {
+    label: 'Long facet shown',
+    name: 'long_facet',
+    values: [
```
dlrice: There are duplicate `value_1` attributes for `value` in this array and because the `id` for the elements sent to the `ExpandableList` are made with `${facet.name}_${value.value}` this is producing duplicate key errors in the console.

xwatkins: Well spotted, just updated it.

PR #39, opened 2019-09-24
Add hero container

xwatkins:
- Adding hero container to display content of News/Help on homepage, but also potentially to use on the entry page instead of Cards.
- Add `small` variant for Tiles

### DIFF
--- a/src/styles/components/tile.scss
+++ b/src/styles/components/tile.scss
@@ -24,10 +24,10 @@ $size: 15rem;
     );
     color: $colour-white;
   }
-  .tile__header {
+  &__header {
     display: block;
   }
-  .tile__content {
+  &__content {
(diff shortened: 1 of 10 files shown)


src/styles/components/tile.scss
```diff
@@ -37,14 +37,14 @@ $size: 15rem;
...
-  &.tile--small {
+  &--small {
     height: $size/4;
     .tile__header {
       font-size: 1.2rem;
     }
   }
 
-  &.tile--uniref {
+  &--uniref {
```
dlrice: Since the `namespace` is one of a limited set in the scss file we can also set the proptypes in Tile.jsx:
```PropTypes.oneOf(['uniref', 'uniparc', 'proteomes'])```

PR #40, opened 2019-09-27
Trm 22038/evidence tags

xwatkins:
Modifying component to fit new requirements.

### DIFF
--- a/src/components/evidence-tag.jsx
+++ b/src/components/evidence-tag.jsx
@@ -1,4 +1,4 @@
-import React, { useState, cloneElement } from 'react';
+import React, { Fragment, useState, cloneElement } from 'react';
 import PropTypes from 'prop-types';
 import Tag from '../svg/tag.svg';
 import '../styles/components/evidence-tag.scss';
@@ -10,7 +10,7 @@ const EvidenceTag = ({
 }) => {
   const [contentDisplay, setContentDisplay] = useState(false);
   return (
-    <span>
+    <Fragment>
       <span
         className={`evidence-tag ${className}`}
         onClick={() => setContentDisplay(!contentDisplay)}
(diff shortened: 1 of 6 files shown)


src/components/evidence-tag.jsx
```diff
@@ -31,12 +50,22 @@ EvidenceTag.propTypes = {
   /**
    * Decides the colour of the tag
    */
-  reviewed: PropTypes.bool,
+  className: PropTypes.string,
+  /**
+   * Decides the colour of the tag
+   */
+  IconComponent: PropTypes.element,
```
dlrice: Getting a PropTypes error when a IconComponent is provided (`IconComponent: ConfigureIcon`). Can either:
1. change PropType to `PropTypes.func`.
2. keep PropType as an element but use `{cloneElement(IconComponent, { width: size, height: size })}` where it's used at L22. Then an element must be passed eg (`IconComponent: < ConfigureIcon />`). Note that the default would need to become `<Tag />`.

src/components/evidence-tag.jsx
```diff
@@ -1,20 +1,39 @@
...
+const EvidenceTag = ({
+  label, title, className, IconComponent, children,
+}) => {
+  const [contentDisplay, setContentDisplay] = useState(false);
   return (
-    <span className={styleClassName}>
-      <Tag width={size} height={size} />
-      <span className="evidence-tag__label" title={title}>
-        {label}
+    <span>
```
dlrice: Swapping `<span>` with `<React.Fragment>` could save a DOM node.

src/components/evidence-tag.jsx
xwatkins: Very nice! Went for 2. and updated the prop name to be `kebabCase`

PR #42, opened 2019-10-02
a few temporary changes to fix the css grid layout issues. 

its-mahdi:
Most of these changes need to be reverted back when implementing a better grid layout system.

### DIFF
--- a/src/styles/components/header.scss
+++ b/src/styles/components/header.scss
@@ -3,6 +3,7 @@
 
 .header,
 .hero-header {
+  grid-area: hero-header;
   display: flex;
   align-items: center;
   align-content: center;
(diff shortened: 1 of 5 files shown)


src/styles/components/header.scss
```diff
@@ -3,6 +3,7 @@
 
 .header,
 .hero-header {
+  grid-area: hero-header;
```
xwatkins: This is causing the header to not display anymore at the top of the application

src/styles/components/tile.scss
```diff
@@ -9,7 +9,7 @@ $size: 15rem;
   position: relative;
   background-color: $colour-uniprot-blue;
   display: inline-block;
-  margin: $global-margin;
+  margin: 0.5rem;
```
xwatkins: Even if this is temporary, could we still use variables? in that case `$global-margin/2`

PR #43, opened 2019-10-04
Trm 22818/annotation score visualisation

dlrice:
* Converting `DoughnutChart`'s `div`s to `span`s (so they can be children of paragraph elements).
* `DoughnutCharts` have `vertical-align: middle;` to replicate `Bubble` alignment.
* Gave `Tile` a `namespace` to silence warning during testing.

### DIFF
--- a/src/styles/components/doughnut-chart.scss
+++ b/src/styles/components/doughnut-chart.scss
@@ -1,6 +1,6 @@
-@import "../settings";
-@import "../colours";
-@import "../common/utils";
+@import '../settings';
+@import '../colours';
+@import '../common/utils';
 
 $border-width: 2px;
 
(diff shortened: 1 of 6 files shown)


src/styles/components/doughnut-chart.scss
```diff
@@ -10,6 +10,7 @@ $border-width: 2px;
   position: relative;
   display: inline-block;
   margin: 0 $global-margin/5;
+  vertical-align: middle;
```
xwatkins: For consistency, should we add this to `bubble.scss` too? If so please check in uniprot-website that it doesn't have unwanted side effects. Thanks!

dlrice: Updated

PR #45, opened 2019-10-15
Add flat option to infolist

xwatkins:
Add "flat" option to InfoList pattern to display only contents.

### DIFF
--- a/src/components/info-list.jsx
+++ b/src/components/info-list.jsx
@@ -1,23 +1,29 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 
 import '../styles/components/info-list.scss';
 
-const InfoList = ({ infoData, columns }) => (
-  <div className={`info-list${columns ? ' info-list--columns' : ''}`}>
-    {infoData.map(
-      // Only draw if there is content
-      item =>
-        item.content && (
-          <div className="info-list__item" key={item.title}>
-            <div className="info-list__item__title">{item.title}</div>
-            <div className="info-list__item__content">{item.content}</div>
-          </div>
-        ),
-    )}
-  </div>
-);
-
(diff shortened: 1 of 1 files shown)


src/components/info-list.jsx
```diff
@@ -1,23 +1,29 @@
...
-            <div className="info-list__item__title">{item.title}</div>
-            <div className="info-list__item__content">{item.content}</div>
-          </div>
-        ),
-    )}
-  </div>
-);
-
+const InfoList = ({ infoData, columns, isFlat }) =>
+  (!isFlat ? (
```
dlrice: Being pedantic here because your logic absolutely correct: am wondering if it would be slightly clearer to condition on `isFlat` rather than `!isFlat`?

dlrice: Being pedantic here because your logic is absolutely correct: am wondering if it would be slightly clearer to condition on `isFlat` rather than `!isFlat`?

src/components/info-list.jsx
```diff
@@ -1,23 +1,29 @@
...
+              <div className="info-list__item__title">{item.title}</div>
+              <div className="info-list__item__content">{item.content}</div>
+            </div>
+          ),
+      )}
+    </div>
+  ) : (
+    <Fragment>
+      {infoData.map((item, index) => (
+        <div key={item.title}>{index === 0 ? <strong>{item.content}</strong> : item.content}</div>
```
dlrice: Seems like the flat mode is not just a different way to view the same `title`/`content` being passed in as the first `content` is now being used like a title and the `title` attributes are no longer displayed. We've said that the purpose of the `InfoList` is to:

> Provide a way of easily scanning for attribute names in order to view their associated data.

Is the first content item now the attribute name and the remaining content items the associated data?

PR #52, opened 2020-01-07
Storybook

xwatkins:
Migrate Franklin to use storybook.
- Create at least 1 story for each component
- Create custom add on for Purpose/Function
- Custom theming
- Create 2 custom decorators (state and data)
- Add markdown for core component document
- Basic categorisation of components to be reviewed later
- Cleanup

To run: `yarn storybook`.


stories/MainSearch.stories.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { action } from '@storybook/addon-actions';
+import MainSearch from '../src/components/main-search';
```
dlrice: I liked how we used to do `import { MainSearch } from '../src/components';` (from the index file) because this also tested that we had exported it properly thus preventing the numerous `cannot find...` errors when attempting to import from uniprot.

stories/DataTable.stories.js
```diff
@@ -0,0 +1,58 @@
...
+}) => (
+  <DataTable
+    {...{
+      data,
+      idKey,
+      columns,
+      hasMoreData,
+      onLoadMoreItems,
+    }}
+    onSelect={action()}
```
dlrice: Could we do something like:

```
export const actionsData = {
  onSelect: action('onSelect'),
  onHeaderClick: action('onHeaderClick'),
};

so that it is displayed which action has been fired when interacting in the storybook?
``

PR #53, opened 2020-01-14
Typography updates

its-mahdi:
Typography updates and colour palette colour overrides. This was a branch off the Modal branch, so ignore the Modal related commits.


src/styles/_colours.scss
```diff
@@ -6,18 +6,18 @@
 @import "./settings";
 
 // PRIMARY COLOURS
-$colour-dark-blue: #004864;
-$colour-uniprot-blue: #006e9a;
-$colour-medium-blue: #3ca3ca;
-$colour-light-blue: #93d4ee;
+$colour-dark-blue: #014371;
```
xwatkins: These need to be renamed to the new names defined by Austra. Also update the colours showcase page.

src/styles/_franklin-settings.scss
```diff
@@ -0,0 +1,102 @@
...
+    'xxlarge': (20.0px, 1.250rem, 1.7, -0.047rem, 'bold')
+  ),
+  'H7': (
+    'small': (20.0px, 1.250 * 0.5rem, 1.7, -0.047rem, 'semi-bold'),
+    'medium': (20.0px, 1.250 * 0.8rem, 1.7, -0.047rem, 'semi-bold'),
+    'large': (20.0px, 1.250rem, 1.7, -0.047rem, 'semi-bold'),
+    'xlarge': (20.0px, 1.250rem, 1.7, -0.047rem, 'semi-bold'),
+    'xxlarge': (20.0px, 1.250rem, 1.7, -0.047rem, 'semi-bold')
+  ),
+  'A1': (
```
xwatkins: I don't think we need to have different font sizing from colour for anchors, it should use the same as normal. I think this stems from confusion with level 3 button.

src/styles/_colours.scss
```diff
@@ -6,18 +6,18 @@
 @import "./settings";
 
 // PRIMARY COLOURS
```
xwatkins: We seem to have lost the background colours for:
- evidence tags (itself but also when clicked to open)
- accordion title
- autocomplete highlight (type "item" then use up and down arrows - compare to current franklin website)
- Hero container
... basically compare with current https://ebi-uniprot.github.io/franklin-sites/ui-components

src/styles/_franklin-settings.scss
```diff
@@ -0,0 +1,102 @@
+// Font families
+$fs-base-font-face: 'Lato', sans-serif;
+$fs-headers-font-face: 'Source Sans Pro', sans-serif;
```
xwatkins: This might be due to something else, but it seems there is an issue with antialiasing of the title font. Tested on chrome but might be other browsers too. Have you added this somewhere:
* {
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
}

src/styles/common/_typography.scss
```diff
@@ -0,0 +1,80 @@
...
+ul {
+  margin-left: 1.25rem;
+  list-style-type: disc;
+}
+ 
+ol {
+  margin-left: 1.25rem;
+}
+
+a, .a1 {
```
xwatkins: See above

src/styles/_colours.scss
its-mahdi: It's done!

src/styles/_franklin-settings.scss
its-mahdi: Done too!

src/styles/_colours.scss
its-mahdi: There is actually a very slight shade of grey there, so nothing technically wrong. The colour code there is as per design.

src/styles/_franklin-settings.scss
its-mahdi: Added the anti-aliasing now!

src/styles/common/_typography.scss
its-mahdi: Done!

src/styles/_colours.scss
xwatkins: Would it be possible to try with one tone darker (azureish-white I think)? At the moment it's impossible to distinguish from the background. Will check with Austra aftwerwards.

its-mahdi: Done!

its-mahdi: done!

PR #56, opened 2020-02-07
Display menu

xwatkins:


src/styles/components/display-menu.scss
```diff
@@ -0,0 +1,32 @@
+@import '../colours';
+@import '../settings';
+
+.display-menu {
+  & > ul > li > h5 {
```
its-mahdi: I have always had the impression that using Child Selector is not a great practice. This is particularly because it reduces the flexibility of the design and can be a lot more prone to bugs when the mark-up is changed. This is also why things like BEM and OOCSS are preferred way of writing CSS.

So, anyway we can get rid of this direct child selectors and write something no entirely bound to child descendant order?

stories/Card.stories.js
```diff
@@ -45,13 +44,11 @@ export const card = () => (
...
-      title="Title"
-      subtitle={<a to="/#">APOE_HUMAN - P02649</a>}
-      links={links}
-    >
-      {getLipsumSentences()}
-    </Card>
-  </Router>
+  <Card
+    title="Title"
+    subtitle={<a to="/#">APOE_HUMAN - P02649</a>}
```
its-mahdi: Shouldn't this be `<Link to="">`, rather `<a to="">`? I think Anchor tag should only be used with `href` attribute.

src/components/__tests__/card.spec.jsx
```diff
@@ -40,28 +30,20 @@ describe('Card component', () => {
...
-        <Router>
-          <Card title="Title" subtitle="Subtitle" onClick={() => {}}>
-            <span>Some content</span>
-          </Card>
-        </Router>,
-      )
-      .toJSON();
-    expect(component).toMatchSnapshot();
+    const { asFragment } = renderWithRouter(
+      <Card title="Title" subtitle="Subtitle" onClick={() => {}}>
```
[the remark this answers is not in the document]
xwatkins: Actually hadn't "written" the tests in this PR, just translated to `react-testing-library` but you're right, I've added it as a proper test now.

src/styles/components/display-menu.scss
xwatkins: True - should be resolved

PR #58, opened 2020-02-26
Trm 23394 sequence

xwatkins:
- Adding dummy buttons for "Add to basket" and "Download"
- Adding links under "Tools" button for expasy tools and blast
- Refactor as functional component

### DIFF
--- a/src/components/sequence.jsx
+++ b/src/components/sequence.jsx
@@ -35,7 +35,14 @@ const sequenceTools = [
   },
 ];
 
-const Sequence = ({ id, sequence, chunkSize, accession, initialTextSize }) => {
+const Sequence = ({
+  id,
+  sequence,
+  chunkSize,
+  accession,
+  initialTextSize,
+  blastPath = '/blast/accession/',
+}) => {
   const [textSize, setTextSize] = useState(initialTextSize);
   const [highlights, setHighlights] = useState([]);
   const [copied, setCopied] = useState(false);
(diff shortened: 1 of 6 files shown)


src/components/sequence.jsx
```diff
@@ -1,31 +1,56 @@
...
+  useEffect(() => {
+    if (!text || initialTextSize) {
       return;
     }
     // Measure height and width of the dummy element
-    const box = this.text.getBBox();
-    this.setState({ textSize: { width: box.width, height: box.height } });
-  }
+    const box = text.current.getBBox();
+    setTextSize({ width: box.width, height: box.height });
```
dlrice: a pedantic/stylistic suggestion:
```
const {width, height} = text.current.getBBox();
setTextSize({width, height});
```

src/components/sequence.jsx
```diff
@@ -81,65 +100,85 @@ class Sequence extends Component {
...
-        </div>
-        <div className="sequence">
-          <div className="sequence__sequence">{content}</div>
+  return (
+    <Fragment>
+      <div className="action-bar button-group">
+        <DropdownButton label="Tools">
+          <ul className="no-bullet">
+            <li>
+              <a href={`/blast/accession/${accession}`}>BLAST</a>
```
dlrice: Do want to give the consumers the option to pass in the path to their blast endpoint?

xwatkins: Good idea! Added

PR #65, opened 2020-03-26
Isoform download url as prop

dlrice:


stories/Sequence.stories.jsx
```diff
@@ -12,6 +12,13 @@ export default {
...
 
+const uniProtSequenceDownload = accession =>
+  `https://wwwdev.ebi.ac.uk/uniprot/api/uniprotkb/accession/${accession}.fasta`;
+
 export const sequence = () => (
-  <Sequence sequence={sequenceData} accession="P05067" />
+  <Sequence
+    sequence={sequenceData}
+    accession="P05067"
+    downloadUrl={uniProtSequenceDownload('P05067')}
```
xwatkins: Might as well just pass the url rather than add that function?

PR #67, opened 2020-03-27
Optional dropdown for main search

its-mahdi:


src/components/__tests__/__snapshots__/main-search.spec.jsx.snap
```diff
@@ -6,65 +6,6 @@ exports[`MainSearch component should render 1`] = `
     class="main-search"
     data-testid="main-search-form"
   >
-    <div
```
dlrice: Would be good to add another test where we take a snapshot of the main-search with a dropdown.

PR #72, opened 2020-04-08
Adding message component

xwatkins:
PR for the "basic message component" - this doesn't include the "popup message"


src/components/message.jsx
```diff
@@ -0,0 +1,50 @@
...
+      </button>
+    )}
+  </section>
+);
+
+Message.propTypes = {
+  /**
+   * The message body
+   */
+  children: PropTypes.element.isRequired,
```
dlrice, line 34: `children` could also be `PropTypes.string`

stories/Message.stories.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { action } from '@storybook/addon-actions';
+import { Message } from '../src/components';
+import { getLipsumSentences } from '../src/mock-data/lipsum';
+
+export default {
+  title: 'Layout|Message',
+  parameters: {
+    purposeFunction: {
+      function: '',
```
dlrice, line 10: Are we leaving this empty until it's more developed?

src/styles/components/message.scss
```diff
@@ -0,0 +1,54 @@
...
+@mixin getWarningCSS($level) {
+  @include SVGDefinition(($level));
+  .alert__content {
+    border-left: getMessageBorder($level);
+  }
+}
+
+.alert {
+  -webkit-box-shadow: 0px 1px 3px 0px rgba(0, 0, 0, 0.2);
+  -moz-box-shadow: 0px 1px 3px 0px rgba(0, 0, 0, 0.2);
```
aurel-l, line 24: Are those prefixes manually written or were they added by a tool? I'd say we needn't worry about those anymore. [see here for compatibility](https://caniuse.com/#feat=mdn-css_properties_box-shadow)

stories/Message.stories.jsx
xwatkins: I forgot! Now added.

src/styles/components/message.scss
xwatkins: Good point - now removed

src/components/message.jsx
aurel-l: `Proptype.node` then?

xwatkins: 👍

src/components/message.jsx
```diff
@@ -0,0 +1,50 @@
...
+  failure: 'failure',
+  success: 'success',
+  info: 'info',
+};
+
+const iconSize = '1.125em';
+
+const Message = ({ children, level, onDismiss }) => (
+  <section className={`alert alert--${level}`}>
+    <section className="alert__content">
```
aurel-l, line 18: Why a section directly inside a section and not a div? Not sure about the semantic meaning here. I'm perfectly happy with the outer section though.

src/components/message.jsx
```diff
@@ -0,0 +1,50 @@
...
+  warning: 'warning',
+  failure: 'failure',
+  success: 'success',
+  info: 'info',
+};
+
+const iconSize = '1.125em';
+
+const Message = ({ children, level, onDismiss }) => (
+  <section className={`alert alert--${level}`}>
```
aurel-l, line 17:
```suggestion
  <section className={`alert alert--${level}`} role="status">
```
For ARIA reasons, [see here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_status_role)

xwatkins:
Think all comments have now been addressed.

PR #80, opened 2020-05-20
added a basic sequence validator function for AA and NA sequences

its-mahdi:


src/validators/__test__/sequenceValidator.spec.js
```diff
@@ -0,0 +1,106 @@
...
+  expectedResult = {
+    ...validResponse,
+    likelyType: 'aa',
+  };
+
+  expect(results)
+    .toEqual(expectedResult);
+});
+
+test('should remove all FASTA description lines', () => {
```
aurel-l, line 97: So it's removing all the comments and considering all of the sequences as one to calculate type likelihood, right?
I don't know if we might rather want to test sequence by sequence, independently, and return an array of likelihood results.

its-mahdi: Yes, that FASTA sample is just basically to try different headers and see if the RegEx works fine on different lines. The requirement doesn't say "handle multiple sequences" and the last time I used recursion, it was removed to make the code simpler. So, I am somewhat reluctant to support multiple sequences, although it doesn't need anything like recursion.

Do you think this function should be able to handle multiple sequences?

aurel-l: I don't know to be honest, because the blast server doesn't handle multiple sequences out of the box anyway, maybe check with the others.
My main point would be that it's doing something quite unexpected to me. Assume I put a short first sequence that would be rejected on its own, but a valid long 2nd sequence. Then this would tell me this is valid (because it's merging both sequences).
In this case I would expect it to either:
 - Test only the first sequence and fail (because too short), and ignore the 2nd
 - Test both sequences and return an array of [<failed>, <ok>]
 - Fail because there is more than one sequence

its-mahdi: Yeah, that makes sense. I will double check that and do either 1st or 3rd from your suggestions.

xwatkins: I think we discussed last week in the design meeting the fact that there was many help desk requests about submitting multiple blast jobs at once, which would require validation of multiple sequences in one go and then sequential submissions. We are not implementing sequential submissions in the first attempt, but it will likely be required in the future.
So maybe suggestions #2 is best

its-mahdi: So in order to keep the function input and return consistent, should I make this function to always accept and array of sequences and always return an array of results too? Only sometimes returning an array can force the user of function to write more code.

aurel-l: That would be sensible. For me OK to return always arrays

its-mahdi: I pushed another commit to handle array of sequences, but it still needs to handle a FASTA string and return an array of results, so please wait until that one is pushed too as well.

aurel-l:
Right now a short sequence without cleanup required (let's say `ACGTUA`) will not be detected as a too-short sequence.

its-mahdi:
Add one test and had to change another one, but it now handles short sequences correctly.

PR #81, opened 2020-05-26
Seq submission component

its-mahdi:
Created a sequence submission component

### DIFF
--- a/src/components/sequence-submission.jsx
+++ b/src/components/sequence-submission.jsx
@@ -1,0 +1,94 @@
+import React, {
+  useState,
+  useEffect,
+  useCallback,
+  useRef,
+  Fragment,
+} from 'react';
+import PropTypes from 'prop-types';
+import validateSequences from '../validators/sequenceValidator';
+import Message from './message';
+import '../styles/components/sequence-submission.scss';
+
+const SequenceSubmission = ({
+  value,
+  onChange,
+  placeholder,
+}) => {
+  const [error, setError] = useState('');
+  const ref = useRef();
+  const autoFocus = () => {
+    ref.current.focus();
+  };
+
+  useEffect(() => autoFocus(), []);
+
+  const onChangeWithValidation = useCallback((newValue) => {
+    // Calling the custom 'onChange' callback first
+    if (onChange instanceof Function) {
+      onChange(newValue);
+    }
+
+    // Reset the error value before starting the validation    
+    setError('');
+
+    // At the moment we are only supporting one sequence, but we need
+    // to send an array to the validator. Here we are sending an array
+    // and extracting the first result object off the returned results.
+    const result = validateSequences([newValue])[0];
+
+    if (!result.valid) {
+      setError(result.message);
+    }
+  }, [onChange]);
+
+  useEffect(() => {
+    if (value) {
+      onChangeWithValidation(value);
+    }
+  }, [value, onChangeWithValidation]);
+
+  return (
+    <Fragment>
+      <textarea
+        className="sequence-submission-input"
+        defaultValue={value}
+        onChange={e => onChangeWithValidation(e.target.value)}
+        placeholder={placeholder}
+        ref={ref}
+        data-testid="sequence-submission-input"
+      />
+      {(error) && (
+        <Message
+          level='failure'
+          data-testid='sequence-submission-error'
+        >
+          {error}
+        </Message>
+      )}
+    </Fragment>
+  );
+};
+
+SequenceSubmission.propTypes = {
+  /**
+   * The pre-loaded value, if needed.
+   */
+  value: PropTypes.string,
+  /**
+   * Triggers when the value is changed.
+   */
+  onChange: PropTypes.func,
+  /**
+   * Display text when the textarea is empty.
+   */
+  placeholder: PropTypes.string,
+};
+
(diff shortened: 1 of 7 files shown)


src/components/sequence-submission.jsx
```diff
@@ -0,0 +1,94 @@
...
+
+const SequenceSubmission = ({
+  value,
+  onChange,
+  placeholder,
+}) => {
+  const [error, setError] = useState('');
+  const ref = useRef();
+  const autoFocus = () => {
+    ref.current.focus();
```
aurel-l, line 21: Not sure about that behaviour by default. I don't like it in general, and it's not really nice for accessibility reasons.
Is there any specific reason why this is here?

its-mahdi: No specific reason. I don't know what accessibility issues the auto-focus can cause, but either try to improve it or remove it altogether.

aurel-l: It's basically the equivalent of being teleported in an unexpected area. A screenreader would jump directly there without any context of where they are within the page.
Also, it might not be the main input for some people as we'll have a different input to load a sequence from an accession.

its-mahdi:
Xav, regarding your points: I didn't realise we would want a submission button on also this component. I thought this is more like a way of entering the sequence. I can add a 'submit' button easily though.

The parent -- or any other component, can receive a copy of sequence -- without validation, by passing the `onChange` method to this component. But if I am going to add a 'submit' button, I will add a local state to keep track of validation status for the submission purposes.

Regarding "sequence is too short", I don't think any real users would start typing the sequence there very often, but I can add an exception there to ignore short sequences from user input.

What I'm gonna do next:
- Add a submit button
- Accept a 'onSubmit' callback. This will take care of actual submission, so needs to be able to make the API calls, etc. If this component should internally handle the API call, then I need to get some details first and figure out what's the best way since this might need some stuff from UniProt website.
- Ignore short sequences until they are submitted

Does that sound logical?
-

xwatkins:
No, I didn't mean add a submit button in the component, I mean the submit button would be in the parent but only enabled when the sequence is valid, which is why I was saying the parent needs to have a way to know whether the sequence is valid or not.

aurel-l:
The onChange could be called with the whole validation object instead of just the sequence

src/components/sequence-submission.jsx
its-mahdi: Okay, perfect, I didn't know that! The auto-focus is now removed!

its-mahdi:
Xav:

> * If we are doing the validation on change instead of on submit I imagine we will want the parent to be aware of whether the sequence is valid or not (for instance to enable the submit button)

This is now added to the response of the `onChange` callback function. The sequence is also returned with the validation result, so the callback owner can initiate the submit or any other thing that requires the user input.

> * Again if we do the validation on change, I don't think the "sequence is too short" should be displayed as you start typing the sequence as it can be annoying

I tried my best to come up with a clean solution for this, but anything I tried looked too hacky. In the end I decided to simply ignore showing *any* errors if the sequence is shorter than 10 characters.

The `onChange` callback will still get the true result of validation though, but no errors will be shown via this component directly.

Let me know if you have any suggestions on how to do this. Whatever I tried was basically adding `if`s and passing extra arguments and either importing text or objects or magically copy-and-pasting them in the code in other files -- e.g. the 'too short' error message itself, which all looked like bad practices.

aurel-l:
So, looking at it more in depth, the inner textarea in an uncontroled element, right? (as in, its value is not coming from state nor from props).
How do you dynamically inject a sequence in there?
Cases:
 - Loading a sequence from an accession/identifier/etc search.
 - Reset form

its-mahdi:
It is an uncontrolled element, but you can change its value dynamically either by passing the `value` prop -- like [this](https://github.com/ebi-uniprot/franklin-sites/blob/seq-submission-component/src/components/__tests__/sequence-submission.spec.jsx#L16), or firing the `change` event -- like [this](https://github.com/ebi-uniprot/franklin-sites/blob/seq-submission-component/src/components/__tests__/sequence-submission.spec.jsx#L43), but the former is probably the best.

I didn't have a need for storing the sequence value in the state, but if there are any use-cases for it, I'm happy to do it.

its-mahdi:
Also now thinking about it, not sure if it is in fact an *uncontrolled* component per se. But the value can be passed in and read out regardless.

aurel-l:
What I mean is something like this, which is probably how we are going to use this component in context.
I added this as a story to be able to try it and it doesn't really work as I expect it to.
```js
export const dynamicallyChangeValue = () => {
  const [sequence, setSequence] = useState('ACTG');
  return (
    <form>
      <SequenceSubmission
        placeholder="Enter a sequence..."
        value={sequence}
        onChange={validationObject => setSequence(validationObject.sequence)}
      />
      <label>
        Sequence:
        <output>{sequence}</output>
      </label>
      <input
        type="reset"
        onClick={e => {
          e.preventDefault();
          setSequence('');
        }}
      />
      <input
        type="button"
        value="async load sequence"
        onClick={() => setSequence('ACTGUACTGUACTGU+')}
      />
    </form>
  );
};
```

its-mahdi:
Ah, sorry about that. I had used `defaultValue` on the textarea instead of `value`. I have added your story and it is working, but please have a look

PR #86, opened 2020-06-04
Sequence submission changes

its-mahdi:
using default value for the component and some styling.


its-mahdi:
You can reset the string at the moment. Can you explain where do you want to add the exceptions for empty strings exactly?

its-mahdi:
I added a simple exception to the if statements that were there for now, but I will get back to this component later on as I suspect it may have another issue as well. For now I need to merge it with master so I can properly check it with the rest of the changes.

PR #88, opened 2020-06-15
Code block

aurel-l:
Create new component to display preformatted code blocks (with light or dark mode available through props)


src/components/code-block.jsx
```diff
@@ -0,0 +1,28 @@
...
+      className ? ` ${className}` : ''
+    }`}
+    {...props}
+  >
+    <code>{children}</code>
+  </pre>
+);
+
+CodeBlock.propTypes = {
+  lightMode: PropTypes.bool,
```
xwatkins, line 18: Adding /** comments */ to proptypes is then picked up by storybook to display the doc. See other components.

aurel-l: Ah, I hadn't realise it was used. I added comments now then.

PR #89, opened 2020-06-15
Modify BEM hierarchy, and define compact modifier

xwatkins:


src/components/data-table.jsx
```diff
@@ -178,11 +176,16 @@ const DataTable = ({
 DataTable.propTypes = {
   ...DataTableHead.propTypes,
   ...DataTableBody.propTypes,
+  /**
+   * Whether the table is compact
+   */
+  compact: PropTypes.bool,
```
aurel-l, line 182: I believe you mentioned you might, eventually, want more levels of "compactness". Might it be more future-proof to have a `density` prop with, at the moment, only 'normal' (default) and 'compact' as accepted values to be able to add more in the future?
They could be exported as:
```js
export const COMPACT = Symbol('COMPACT');
export const NORMAL = Symbol('NORMAL');
```
(see [this article](https://spicefactory.co/blog/2019/03/26/how-to-avoid-the-boolean-trap-when-designing-react-components/))

src/components/data-table.jsx
```diff
@@ -155,9 +152,10 @@ const DataTable = ({
   getIdKey,
   selectable,
   onHeaderClick,
+  compact,
 }) => (
   <Fragment>
```
aurel-l, line 157: Why is there a Fragment here, if this is rendering only one child at the root of this component?

src/components/data-table.jsx
```diff
@@ -155,9 +152,10 @@ const DataTable = ({
   getIdKey,
   selectable,
   onHeaderClick,
+  compact,
```
aurel-l, line 155: It would be good to have a `...propsForTable` to be able to pass any other prop to the underlying `<table>` tag.
It would need to merge a possible className prop like so:
```jsx
<table className={`data-table${compact ? ' data-table--compact' : ''}${propsForTable.className ? ` ${propsForTable.className}` : ''}`}>
```

xwatkins:
Valid comments, all addressed.

PR #90, opened 2020-06-16
Change tabs

aurel-l:
- Change Tabs logic to only render the content that is visible (instead of hiding invisible content)
 - Change Tabs API to use combination of `<Tabs>` and `<Tab>` within
 - Change structure and attributes to get closer to [ARIA recommendations](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html) regarding tabs (not 100% following yet though)
 - Have Tabs automatically manage their own state, or be managed through props (`active` prop on `<Tabs>` component)
 - More examples in storybook
 - Changed 2 linting rules as they were clashing with automatic code formatting.


src/components/tabs.jsx
```diff
@@ -1,82 +1,141 @@
...
-      data-testid="tab-title"
-      role="button"
-      tabIndex={0}
-      onClick={onClick}
-      onKeyPress={key => handleKeyPress(key)}
-      className={`tabs__header__item ${active ? 'tabs__header__item--active' : ''}`}
-    >
-      {children}
-    </div>
+    const defaultSelected = tabs.find(tab => tab.defaultSelected);
```
xwatkins, line 55: As it's not impossible to accidentally specify `defaultSelected` on more than one tabs, I wonder if it might be worth either throwing an error or logging it in the console if it's the case.

aurel-l: A warning has now been added

PR #91, opened 2020-06-16
improved styling. flexbox -> grid. bug fixes.

its-mahdi:
This was requested by Austra to match the newer designs from the mock-ups with the message component. Here is a summary of what is done in this PR:

- There was a bug here where a long line of text would shrink the message icon on the left-side of the text to a point that it was almost invisible. I fixed it first using flexbox settings, but then completely moved to grids.
- The icon on the left needed better margin settings.
- The dismiss button on the right was vertically aligned in the centre. This mean multiple lines of content in the message itself would make the dismiss icon look a bit odd. This is fixed now.
- Adding the second line of content -- as per mock-ups, using flexbox was a little bit tricky, therefore switched to the grid.
- Added an option to hide the default icon, whenever needed -- to match the mock-ups.
- Had to create a dummy element to represent the side border on the left side of the message. This was due to using grid, optional icon and a few other cases would make the original `border-left` approach very tricky.
- Used `h3` for the actual message content to match the mock-ups. This inherits some responsive behaviour, just a heads-up.
- Had to add a `no-icon` modifier to accommodate the lack of icon in the grid settings.
- The second line is called `subtitle`, but happy to change it to a better name if anyone has a better suggestion.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "react-router-dom": "^5.0.1"
   },
   "dependencies": {
+    "classnames": "^2.2.6",
     "file-loader": "^2.0.0",
     "foundation-sites": "^6.5.3",
     "history": "^4.10.1",
--- a/src/components/__tests__/__snapshots__/message.spec.jsx.snap
+++ b/src/components/__tests__/__snapshots__/message.spec.jsx.snap
@@ -3,21 +3,24 @@
 exports[`Message component should render 1`] = `
 <DocumentFragment>
   <div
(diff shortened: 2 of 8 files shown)


aurel-l:
I just updated with changes for the icon.
(Sorry I had to force push, something went wront when I cancel my previous push)

its-mahdi:
Great, thanks! I had just fixed it, but failed to push it because of your commit. But your fix is more correct than mine, so I'm happy with it too!

PR #95, opened 2020-06-23
Compact chip

xwatkins:


src/components/chip.jsx
```diff
@@ -3,8 +3,13 @@ import PropTypes from 'prop-types';
 import RemoveIcon from '../svg/times.svg';
 import '../styles/components/chip.scss';
 
-const Chip = ({ children, onRemove, className, disabled }) => (
-  <span className={`chip ${disabled ? 'chip--disabled' : ''} ${className}`}>
+const Chip = ({ children, onRemove, className, disabled, compact, title }) => (
+  <span
+    className={`chip ${disabled ? 'chip--disabled' : ''} ${
```
aurel-l, line 8: Maybe we can start using the `classnames` library now that Mahdi added it somewhere else in the codebase. My OCD is getting triggered when I see the 3 white spaces in the generated class in the corresponding tests 😅

xwatkins: The order of the world is now balanced 🧘‍♂️

aurel-l: you didn't push yet, though?

aurel-l: 🧘‍♂️

PR #102, opened 2020-09-09
Data loader provide parent ref to observe scroll trm 24703

dlrice:
Relevant jira: https://www.ebi.ac.uk/panda/jira/browse/TRM-24703

Required refactoring the data loader to accept a reference (eg found with a query selector) as a parent in the uniprot website requires observing rather than a child.


src/components/data-loader.jsx
```diff
@@ -36,28 +52,33 @@ const withDataLoader = BaseComponent => (props) => {
...
-  };
-
-  const checkLoadMoreItems = () => {
-    if (!loading && hasMoreData && isBottom(ref)) {
-      setLoadMoreItems(true);
-    }
-  };
-
+  useEffect(() => {
+    const ref = scrollRef.current;
```
xwatkins, line 56: Wondering whether you need to check if `ref` is not null before continuing?

src/styles/components/data-table.scss
```diff
@@ -26,7 +26,7 @@
     color: $colour-sky-white;
     background-color: $colour-sea-blue;
     top: 0;
-    z-index: 2;
+    z-index: 1;
```
xwatkins, line 29: I think there's some scss variables defined somewhere for `z-index`. It often gets messy so using the variables can help ensure where things are in relation to each other in the UI.

aurel-l (approved):
Given @xwatkins 's comments, once those are fixed I'd consider it approved from my side

src/styles/components/data-table.scss
dlrice: good point. done

src/components/data-loader.jsx
dlrice: good point. done

PR #105, opened 2020-09-23
Add href option to header

xwatkins:


src/__tests__/expectsToThrow.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable no-console */
+const expectToThrow = (func) => {
```
aurel-l, line 2: Just wondering what's the point of that one? Is it just to catch the console error and avoid displaying it?
If so, I'd make the name of that a bit more descriptive

stories/Header.stories.jsx
```diff
@@ -31,7 +32,7 @@ const Search = () => {
   const [value, setValue] = useState('');
   return (
     <MainSearch
-      onChange={v => setValue(v)}
+      onChange={(v) => setValue(v)}
```
aurel-l, line 35: `onChange={setValue}` should be enough here

src/__tests__/expectsToThrow.js
xwatkins: Yes, to avoid seeing a big red output when it's actually expected... I thought the name was good, bu added the "NoConsole" suffix and moved it to `testUtils`.

aurel-l (approved):

Issue #121, opened 2020-11-03
Dropdown render optimisation


aurel-l:
Don't render child of Dropdown if not open.

Might help with some tests, because they were getting elements within the closed dropdown.
Related to #120 as if it renders only on open it makes it easier to make it autofocusable without having to know the open/closed status

aurel-l:
handled in #122

Issue #120, opened 2020-11-03
Tree select text field autofocus


aurel-l:
Have tree select autofocus on open if an autocomplete text field is present.

For simplicity, have it autofocus on render just with the `autofocus` attribute (requires #121)

aurel-l:
handled in #122

Issue #118, opened 2020-10-27
Colour Pumpkin 


AustraAnna:
**Name of component**
New colour for UniRef section, "Pumpkin" F2994C

**Additional modifiers** 
Subsection in colours "Name space" update  - colour for UniRef "Pumpkin"  F2994C

**Description of the component**
"Pumpkin" F2994C used in UniRef header or related designs to indicate UniRef

**Description of the behaviour (including modifiers)**
Colour - needed or selected components are altered to "Pumpkin" once user is on UniRef pages

**Function**
Indicate the user that is landed on UniRef pages

**Purpose** 
Cohesive style element from home page to UniRef landing page

**Add any relevant screenshots/mockups**
![Screenshot 2020-10-27 at 15 07 58](https://user-images.githubusercontent.com/72505072/97320973-49dec880-1866-11eb-96f8-b769d00c6e8e.png)
![Screenshot 2020-10-27 at 15 08 15](https://user-images.githubusercontent.com/72505072/97320987-4d724f80-1866-11eb-8941-dabc6b8d1f5e.png)
![Screenshot 2020-10-27 at 15 07 43](https://user-images.githubusercontent.com/72505072/97321001-506d4000-1866-11eb-919f-d1907a44bb94.png)

aurel-l:
handled in #122

PR #122, opened 2020-11-09
Uniparc icon, and more...

aurel-l:
- Change colour for UniRef #118 
 - Change colour for UniParc #123 
 - Add icon for UniParc
 - Optimise Dropdown by not rendering content until open #121 
 - TreeSelect autofocus on open if it has a search field #120 
   - needed previous changes in Dropdown
   - maybe not the best approach, as it's declarative and not imperative (`autofocus` attribute, instead of calling the `focus()` method through refs, but too many component levels to go through)
 - Fix accessibility issue with Dropdown (Invisible focus)
 - Fix accessibility issue in website header (could reach fake "dropdowns")
   - now uses actual Dropdown component
 - Match uniprot-website configs
   - TypeScript support
     - switched Tile component to TS to try it out
     - switched entry point from `index.js` to `index.ts`


src/components/dropdown-button.jsx
```diff
@@ -54,7 +54,10 @@ const DropdownButton = ({ children, label, className }) => {
   }, [size]);
 
   return (
-    <div className="dropdown-container">
+    <div
```
xwatkins, line 57: Noticed that the carets are not visible for level 2 and 3 buttons. Also the button margin has changed, don't know if that was on purpose or not
image
image

src/components/header.jsx
```diff
@@ -11,9 +15,16 @@ const HeaderLink = ({ link }) => {
   return (
     <>
       {link.path ? (
-        <Link to={link.path}>{link.label}</Link>
+        <Link to={link.path} className="button tertiary">
```
xwatkins, line 18: The difference between the header "dropdown" and a regular dropdown button is that it should display on mouseover, not on click.

xwatkins (changes requested):

src/components/header.jsx
aurel-l: The issue is that it wasn't displaying on focus. I'll make it so that it appears on mouseover AND on focus then

src/components/dropdown-button.jsx
aurel-l: Fixed the arrow colour.
I removed the margin because I didn't know why it was there and we had to fight against it in the places where we want to use it anyway (for example, the query build alignment issue was mainly that, and now we have a `margin-right: 0` against it)

aurel-l: bottom margin back now, and overriding it in the header

xwatkins (approved):

PR #124, opened 2020-11-09
Other key for facet

aurel-l:
Ability to use an other key name for the facets, apart from the default 'facets'.

Useful for the UniRef entry page, where the members filter visually looks like a facet (so we would use the Facets component), but URL-wise, it actually uses 'filter'. The logic is the same though (active state, selection).

for an example of where it would be used, see the current website: [https://www.uniprot.org/uniref/UniRef50_P61887?filter=reviewed:yes](https://www.uniprot.org/uniref/UniRef50_P61887?filter=reviewed:yes)


src/components/facets.jsx
```diff
@@ -38,58 +38,64 @@ const parse = string => {
  * and generate a search string
  * @param {object} parsedSearch
  */
-const stringify = ({ facets = {}, ...rest }) => {
+export const stringify = (query, queryStringKey = 'facets') => {
```
dlrice, line 41: `queryStringKey = 'facets'` already listed in `Facets.defaultProps`

dlrice (approved):

src/components/facets.jsx
aurel-l: This is for the helper function, whereas the default prop is for the component. Just in case we end up using the function on its own outside of the component

PR #129, opened 2020-11-17
Hero header changes

its-mahdi:
Ticket: https://www.ebi.ac.uk/panda/jira/browse/TRM-24869

### DIFF
--- a/src/styles/components/header.scss
+++ b/src/styles/components/header.scss
@@ -2,6 +2,8 @@
 @import '../colours';
 @import '../mixins';
 
+$content-width: 90%;
+
 .header,
 .hero-header {
   display: flex;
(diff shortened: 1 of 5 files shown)


src/styles/components/header.scss
```diff
@@ -117,7 +117,15 @@
   flex-direction: column;
   margin-bottom: 2em;
 
-  .hero-header__title {
+  &__top {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    flex: 2 1 auto;
+    width: 90%;
```
xwatkins, line 125: Maybe make the width a variable and use in both places where it's 90%.

xwatkins (approved):
Just a minor comment but all good otherwise

PR #132, opened 2020-11-25
Refactor data loader

aurel-l:
Mainly address the issue of the data-loader wrapper depending on scroll position to load more data:
 - -> Problematic when the container was not scrolling (just the document)
 - Now depends on visibility of the bottom of the list (through IntersectionObserver)
 - Note that we don't need a "scrollDataAttribute" anymore

Updated the stories and added one to showcase the new option of "click to load more" either when it is toggled by the user of the component, or because the final user's browser doesn't support IntersectionObserver.

A lot of changes to tests here and there just to keep the coverage at the same level. Some updates to TypeScript too.


src/components/data-table.jsx
```diff
@@ -69,6 +69,7 @@ const DataTableHead = ({ selectable, columns, onHeaderClick }) => (
             key={name}
             className={className}
             onClick={onClick}
+            data-testid={sortable ? 'sortable-header' : undefined}
```
dlrice, line 72: Just wondering why not just call it something like `header-cell` rather than conditioning?

dlrice (approved):

package.json
```diff
@@ -14,6 +14,7 @@
   "scripts": {
     "build": "yarn build-library && yarn build-storybook",
     "build-library": "webpack",
+    "preanalyse": "yarn run build-library",
```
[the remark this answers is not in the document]
aurel-l: Not really related to this PR, but I ended up running "analyse" on an out-of-date build so was seeing the wrong packages in the bundle analysis. So this is to ensure yarn runs the process to build franklin before asking to analyse the bundle

src/components/data-table.jsx
aurel-l: Good point, I guess I was just trying to replicate the previous specific class selector and ended up still being over-specific. I replaced it now in the test and removed the requirement for a testid

PR #137, opened 2020-12-01
Modify tile look and behaviour, update tests

xwatkins:


src/components/tile.tsx
```diff
@@ -1,45 +1,98 @@
...
-    <div className={classNames}>
-      <h3 className="tile__header">{title}</h3>
-      {!small && <p className="tile__content">{description}</p>}
+    <div ref={containerRef}>
+      <div
+        className={classes}
+        style={
+          {
+            '--tile-background': backgroundColor,
+            width: size,
```
aurel-l, line 76: Was there an issue directly using `width: sideLength || sizes.width`?

src/components/tile.tsx
```diff
@@ -1,45 +1,98 @@
...
+            height: size,
+          } as CSSProperties
+        }
+        onClick={onClick}
+        role="button"
```
aurel-l, line 81: I'd rather use a proper button element here instead of having to add role, tabIndex, and having to handle onKeyPress (if it's a button, onClick will handle everything)

src/styles/components/tile.scss
```diff
@@ -1,82 +1,83 @@
...
-    color: $colour-sky-white;
-  }
+
   &__header {
     display: block;
   }
-  &__content {
-    padding: 0.7em;
+  &__description {
+    transition: height $animation-time linear;
```
aurel-l, line 33: Don't transition height, it's expensive to transition (https://csstriggers.com/height). Transition transform instead (https://csstriggers.com/transform)

src/styles/components/tile.scss
```diff
@@ -1,82 +1,83 @@
...
 
-  &--uniref {
-    background-color: $colour-uniref !important;
-    &:hover {
-      background-color: scale-color(
-        $colour-uniref,
-        $lightness: $hover-lightness-value
-      ) !important;
+  // Animations
+  &:hover {
```
aurel-l, line 62:
```suggestion
  &:hover, &:focus, &:focus-within {
```

aurel-l (changes requested):
left a few comment here and there that need to be addressed, apart from that 👍

src/components/tile.tsx
xwatkins: `sizes` is `null` upon first render and it's not triggering a re-render when it updates. Maybe there's another way to handle this?

aurel-l: It is supposed to trigger a re-render, sizes is stored inside the hook as a react useState. Unless there's a bug and it is mutated, it should trigger re-render.

That's one thing. the other thing is the need to measure the container to define the size of this. I guess a good approach need't use this and would just use CSS to fill the available space and keep it a square.

xwatkins:
Implemented changes

package.json
```diff
@@ -35,7 +35,7 @@
     "IE 11"
   ],
   "jest": {
-    "coverageReporters": [
+    " geReporters": [
```
aurel-l, line 38: Not sure that was wanted

aurel-l (approved):

PR #139, opened 2020-12-08
Colour adjustments

xwatkins:
- Changes to the tile and colours following discussions with Austra
- HeroHeader -> tsx


its-mahdi (approved):

src/components/__tests__/__snapshots__/hero-header.spec.jsx.snap
```diff
@@ -3,21 +3,17 @@
 exports[`Hero Header component should render 1`] = `
 <div
   className="hero-header"
+  topLeft={
```
aurel-l, line 6: I guess here the test itself needs to be reviewed rather than the snapshot

src/components/hero-header.tsx
```diff
@@ -0,0 +1,29 @@
...
+  footer?: JSX.Element;
+  className?: string;
+};
+
+const HeroHeader: React.FC<HeroHeaderProps> = ({
+  title = '',
+  children,
+  footer,
+  className,
+  ...rest
```
aurel-l, line 18: After checking usage I'd call it `props` instead of `rest` as it's more consistent with the rest of the codebase

aurel-l (changes requested):
The snapshot looks incorrect

aurel-l (approved):

PR #142, opened 2020-12-10
Grid system

xwatkins:
- [ ] What is the link to the Jira that this PR is for?
- [ ] Have you written tests or updated tests so that the overall coverage doesn't fall?
- [ ] Are there any linting errors that required you to add exceptions?
- [ ] Is there a story associated with it? If so:
   - [ ] Which one(s) should the reviewer look at to evaluate the changes?
   - [ ] Is the automatic documentation in the story up-to-date?
   - [ ] Are the props modifiables in the story through knobs? Or are there multiple stories covering enough different cases?


src/components/header.jsx
```diff
@@ -50,8 +50,11 @@ HeaderLink.propTypes = {
   link: linkProp.isRequired,
 };
 
-const Header = ({ logo, links, search, isNegative }) => (
-  <div className={cn('header', { 'header--negative': isNegative })}>
+const Header = ({ logo, links, search, isNegative, className, ...rest }) => (
+  <div
+    className={cn('header', { 'header--negative': isNegative }, className)}
+    {...rest}
```
xwatkins, line 56: for consistency, we're calling `rest` `props`

src/styles/common/_grid.scss
```diff
@@ -0,0 +1,54 @@
...
+      width: max-content;
+      justify-self: center;
+      // span from side to side
+      grid-column: 1 / -1;
+    }
+  }
+}
+
+.grid-content-block {
+  margin: 1em 0;
```
xwatkins, line 53: You can use `$global-margin` here

src/styles/common/_grid.scss
```diff
@@ -0,0 +1,54 @@
+$gap: 1rem;
+$max-width: 100ch;
+$min-side-margin: 1rem;
+$grid-gap: 1em;
+$separator-width: 2px;
```
xwatkins, line 5: Use `rem` instead of `px` (valid for this line and the next)

src/styles/common/_grid.scss
```diff
@@ -0,0 +1,54 @@
+$gap: 1rem;
+$max-width: 100ch;
```
xwatkins, line 2: How was this number chosen?
Is there a good reason for using `ch` instead of `rem`?

src/styles/common/_grid.scss
```diff
@@ -0,0 +1,54 @@
...
+        // left-margin
+
+    minmax($min-side-margin, calc((100% - #{$max-width}) / 2))
+    // main grid columns
+    repeat(12, 1fr)
+    // right-margin
+    minmax($min-side-margin, calc((100% - #{$max-width}) / 2));
+
+  &--with-background > * {
+    background: salmon;
```
xwatkins, line 22: I guess this is not part of the grid and should be handled separately?

src/styles/common/_grid.scss
aurel-l: The unit was chosen because it matches best what we want to express when we decide the main content width (as in UX says "The main content should be x characters wide" -> `max-width: <x>ch;`).
The value itself I just chose a simple value when doing the prototype.
We need to discuss it with UX looking at recommendations for accessibility. We probably need to consider different main-content widths depending on the type of content (text-heavy as an entry page _vs_ other as the home page)

src/styles/common/_grid.scss
its-mahdi: Where do you think is a better place to put it?

its-mahdi:
Another update here with a few minor fixes, classes to offset the grid cells from left and a story to demonstrate the grid too.

A couple of issues here:
- There is a chance the grid needs some changes to make it work with multi-row alignments. The issue being the `:first-child` selector we use to move the first item one step ahead (first cell is the gutter), won't work with multiple-rows for clear reasons.
- At the moment to solve a CSS specificity issues and conflict all grid cells need to have a `uniprot-grid-cell` class name, however, this gets on the way when we want to use offset classes. So, offset cell should not have the `uniprot-grid-cell` class assigned to them. This is pretty bad and is going to cause a lot of confusion and I'm still trying to work this out, but please have a look and let me know if you have any suggestions.

src/styles/common/_grid.scss
```diff
@@ -0,0 +1,60 @@
...
+  grid-template-columns:
+        // left-margin
+
+    minmax($min-side-margin, calc((100% - #{$max-width}) / 2))
+    // main grid columns
+    repeat(12, 1fr)
+    // right-margin
+    minmax($min-side-margin, calc((100% - #{$max-width}) / 2));
+
+  & > .uniprot-grid-cell:first-child {
```
aurel-l, line 24: How about something like this to avoid having this rule taking over in the case we explicitely set on of the offset classes?
```suggestion
  & > .uniprot-grid-cell:not([class^="uniprot-grid-cell--offset-"]):first-child {
```

its-mahdi: This is the type of CSS nightmare I have always been afraid of! Let me try and thanks for the suggestion!

aurel-l: I'd say you might be able to remove the need for the `uniprot-grid-cell` class name altogether then, as no offset class would default to this first-child rule, and adding a offset class would be required for any complex grid system with 2 or more rows

its-mahdi: This only works if the offset class -- the one that we are excluding, is the first in the `calssName` attribute. I'm trying to find a way now to search the whole `className` rather only checking the start.

aurel-l: Just drop the magic then and make the use of an offset class required 🤷🏽‍♂️

PR #151, opened 2020-12-22
Convert jsx to tsx. If user specifies a boolean for disabled consider…

dlrice:
… this regardless of what onClick is.

## Purpose
* Needed to modify the `disabled` prop behaviour to be used in https://github.com/ebi-uniprot/uniprot-website/pull/240.
* Converted to typescript.

## Approach
When consumer specifically sets `disabled` prop with a boolean, this will be respected regardless of `onClick`'s value.

## Testing
Add tests for presence of disabled prop.

## Stories
Didn't change stories.

## Checklist


src/components/__tests__/chip.spec.jsx
```diff
@@ -23,4 +23,14 @@ describe('Chip component', () => {
     const removeIcon = queryByTestId('remove-icon');
     expect(removeIcon).toBeNull();
   });
+
+  it('should not have disabled class when disabled=false passed as a prop', () => {
+    const { container } = render(<Chip disabled={false}>Some content</Chip>);
+    expect(container.firstChild.className).not.toMatch(/disabled/);
```
aurel-l, line 29: I would probably prefer to test the actual `disabled` attribute instead of the class name as it's more representative of what the browser will allow the user to do.
And, to make it more robust to possible implementation changes, I would go for a testing-library query (...ByTestID, ...ByRole) instead of getting the firstChild.
Something like
```js
expect(screen.getByRole('button')).toHaveAttribute('disabled')
```

dlrice: Updated

aurel-l (approved):

PR #153, opened 2021-01-05
info-list line in its own component

its-mahdi:
Going with the requirements for the supporting data section in the home page, this turned out to be a little bit tricky for a single component to be used both in the InfoList and also independently in the supporting data section. Please have a look at the changes and comment if you have any suggestions.

## Purpose
Creating a reusable element to use both inside `InfoList` element and in other places e.g. Supporting Data section in the home page.

## Approach
Created a new standalone element called `DecoratedListItem` which is used both in `InfoList` and can be used in other places on its own.

## Testing
New snapshots for the new component and updated snapshots as the class name `info-list-item` and it's variations doesn't exist anymore.

## Stories
Just a single story to showcase how the alternative styling is achieved in the Supporting Data section of the home page. All other styles are demonstrated in the `InfoList` stories which needed no updates.

## Checklist
   New component is TSX, however the existing `InfoList` which is very slightly modified is still JSX, due to having other priorities at the moment.
   Not yet.


its-mahdi:
To explain better what was the issue here: The goal was to use one component to create the styled elements in the Supporting Data section of the home page as well as the items that are in the InfoList component, since they both look similar.

The first logical step was to take out the mark-up from the InfoList component which is responsible for creating the styled lines. They have a title in bold, a decorative separator line and content. The title can be optional and if there is no content, nothing will be shown. The InfoList also has other features like making the first line of the list in bold -- highlight property, or make them look more compact for smaller screens.

If the code responsible for this part can be taken out, it can be used in both Supporting Data section and the InfoList. After making a component called InfoListItem, I realised some features in the InfoList component like highlight and compact use a BEM style classes which require the parent to have `info-list` in the className, but also the child which is now `info-list-item` also to uses those `info-list` base class names, rather `info-list-item`.

I decided an approach like that can be confusing in the future, so moved those styling logic to the `info-list-item`. The second issue came about when I realised the decorative line is part of the `title` of each item, which means the line is placed on the right side of the title. This would not work for Supporting Data section as the titles there need to have the decorative line on the left.

Instead of adding an option to switch the line from left to right side of the title, I decided to move the CSS code to the `content` section of the `Info-list-item`. The content is on the left side so it can have the decorative line on the left -- which is same as having the line on the right side of the title, but it can also be used in the Supporting Data section.

In the Supporting Data section I started using the `content` section of the `Info-list-item` instead of the title and it worked fine. Now, under the Supporting Data text, we have two elements that share the same grid cell, but the `info-list-item` takes up 100% of the space available. In order to make these two cells split up the space, I had to give the grid cell `display: inline-flex;` and pass a custom class to `display-info-list`, which is not ideal, because you may think the `info-list-item--inline` will work on its own, but in reality it needs the parent to have `display: inline-flex`. Now that the BEM scope of `info-list` and `info-list-item` are different, that can't be achieved without manually setting proper class modifier on both parent and child.

And after that, it turned out that the styling of the decorative line we use in the Supporting Data section is different than the one we use in the InfoList -- as per mock-ups, so we have to wait and talk to Austra to see if that's okay and which styling should be used in the end.

src/components/info-list-item.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+const InfoListItem: FC<Props> = ({
+  title,
+  content,
+  children,
+  highlight,
+  compact,
+  hideTitle,
+  inline,
+}: Props) => (
```
aurel-l, line 51: This is not needed as you already have `FC<Props>` higher

src/components/info-list-item.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+const InfoListItem: FC<Props> = ({
+  title,
+  content,
+  children,
+  highlight,
+  compact,
+  hideTitle,
+  inline,
```
aurel-l, line 50: Please add `className` and `...props` as per other components in Franklin so that we can pass classes and other props from higher to the underlying `div`.
Not sure if that might actually help you with passing classes from the `InfoList` components and solve some of the inconsistencies you noticed you had to do when extracting this component as standalone

src/components/info-list.jsx
```diff
@@ -10,31 +11,19 @@ const InfoList = ({
...
 }) => (
-  <div
-    className={classNames(
-      'info-list',
-      { 'info-list--columns': columns },
-      { 'info-list--compact': isCompact },
-      { 'info-list--no-title': noTitles }
-    )}
-  >
+  <div className={classNames('info-list', { 'info-list--columns': columns })}>
```
aurel-l, line 14: This should probably be a `ul` and the `InfoListItem` in the map should probably be wrapped in a `li`

src/components/info-list.jsx
```diff
@@ -10,31 +11,19 @@ const InfoList = ({
   highlightFirstItem,
   noTitles,
```
aurel-l, line 12: Same thing here, add `className` and `...props`

src/components/info-list-item.tsx
its-mahdi: done!

src/components/info-list-item.tsx
its-mahdi: done!

src/components/info-list.jsx
its-mahdi: I decided not to go for this because now `DecoratedListItem` is a standalone element and can be used in its on context e.g. we use `DecoratedListItem` in a grid-cell alone in the Supporting Data section in the home page.

src/components/info-list.jsx
its-mahdi: done!

src/components/info-list.jsx
aurel-l: That's why I suggested putting both `<ul>` and `<li>` markup in the `InfoList` element. I meant it like so:
```ts
<ul className={classNames('info-list', { 'info-list--columns': columns })}>
    {infoData.map(
      // Only draw if there is content
      (item, index) =>
        item.content && (
          <li key={item.title}>
            <InfoListItem
              title={item.title}
              content={item.content}
              highlight={index === 0 && highlightFirstItem}
              compact={isCompact}
              hideTitle={noTitles}
            />
          />
        )
    )}
  </ul>
```
I guess it's not a deal breaker though

aurel-l (approved):

src/components/info-list.jsx
its-mahdi: Oh, sorry, I misunderstood this one. I will do it now!

PR #154, opened 2021-01-05
Color refactor

xwatkins:
## Purpose
Even though we'd moved to a single source of truth for colour definition (in a json file), some of the colours had to be redefined in SASS. There was also a mix of colour names describing the colour and some describing their purpose (namespace colours etc...).

## Approach
Removed the SASS colour definition and turned the JSON file in to a js file, where colours are first defined according to what they look like then according to their purpose.
As the SASS `lighten` mixin was used I had to recreate a similar functionality in JS - by creating a function which converts HEX to RGBA (optional transparency value).

## Testing
As the json to scss module only works with module.exports, no tests were added to check the use of the HEX to RGBA convertion and the coverage had to be slightly reduced.

## Stories
What story(ies) did you write or change to document the functionality / changes?
N/A

## Checklist


src/styles/colours.js
```diff
@@ -0,0 +1,78 @@
...
+  colourAmethyst: '#A748BD',
+  colourCalypso: '#357B92',
+  colourCerulean: '#00A6D5',
+  colourTurmeric: '#B8CE48',
+  colourSunglow: '#ffcc33',
+  colourFroly: '#f36968',
+  colourSalem: '#108f3b',
+  colourMalibu: '#79cbf8',
+  colourCoyoteBrown: '#966336',
+  colourOuterSpace: '#374343',
```
its-mahdi, line 35: Is hungarian notation necessary here? e.g. prefixing everything with `colour`? This is generally not recommended, so it will be good if we can get rid of it.

src/styles/colours.js
```diff
@@ -0,0 +1,78 @@
...
+  colourIdMapping: colorDefinition.colourCalypso,
+  colourBlast: colorDefinition.colourCerulean,
+  colourAlign: colorDefinition.colourTurmeric,
+  colourWarning: colorDefinition.colourSunglow,
+  colourFailure: colorDefinition.colourFroly,
+  colourSuccess: colorDefinition.colourSalem,
+  colourInfo: colorDefinition.colourMalibu,
+  colourLink: colorDefinition.colourSapphireBlue,
+  colourSelected: colorDefinition.colourGainsborough,
+  colourHover: hexToRGBA(colorDefinition.colourPlatinum, 0.05),
```
its-mahdi, line 77: Same again with hungarian notation.

its-mahdi (commented):
My main objection would be to start using an existing package for dealing with colours in JS, rather writing our own functions. e.g.

https://www.npmjs.com/package/color

src/styles/colours.js
xwatkins: About the use of Hungarian notation for these variables: as these are used globally as css variables, I think it's important to keep the "colour" prefix. `$colour-independence` is more meaningful than `$independence`.

src/styles/_mixins.scss
```diff
@@ -67,3 +67,12 @@ $global-right: if($global-text-direction == rtl, left, right);
   overflow: hidden;
   text-overflow: ellipsis;
 }
+
+@mixin hover($with-transition: false) {
+  filter: brightness(110%);
+  @if $with-transition == true {
+    transition: 0.5s filter ease;
+  } @else {
+    transition: none;
```
aurel-l, line 76: It would be better to not define the else altogether, as it's not needed and might deactivate other transitions on the elements using the hover mixin

aurel-l (changes requested):
Just one comment about an extra `@else` that isn't needed, otherwise lgtm

src/styles/_mixins.scss
xwatkins: I added it because the `a:hover` was applied to some of the components where we didn't want it to be applied (like in facets). Maybe there's a better way to handle this?

aurel-l: OK, that's strange and probably an issue that it would do that in facets if we don't want it too. But I guess the transition override shouldn't bother us too much anyway in other places.
Go for it then, you can keep it

aurel-l (approved):

PR #155, opened 2021-01-06
Header changes

aurel-l:
## Purpose
Changes to the header logic to accept buttons, needed to trigger opening of the side panel from the header.

## Approach
Recognise passing an `onClick` prop as requiring a button to be rendered instead of a link

## Testing
Snapshot testing to handle more combinations of cases

## Stories
Modified existing story

## Checklist

Also:
 - Switched to using the term `items` instead of `links`.
 - Fixed storybook actions in some places


src/components/header.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+import DropdownButton from './dropdown-button';
+import Button from './button';
+
+import '../styles/components/header.scss';
+
+type HeaderExternalLink = {
+  label: ReactNode;
+  href: string;
+  path?: never;
```
dlrice, line 13: I wasn't sure what `never` was for so I had a google and came across [this article](https://blog.logrocket.com/when-to-use-never-and-unknown-in-typescript-5e4d6c5799ad/) that summarises:

> * Use never in positions where there will not or should not be a value.
> * Use unknown where there will be a value, but it might have any type.
> * Avoid using any unless you really need an unsafe escape hatch.

Given that, it looks like these might be better as `unknown`?

aurel-l: This optional `never` here is to say that if there is an `href`, I don't want there to be a `path`, and you'll find that lower you have the opposite case. This, combined with the `|` lower, mean that I want to accept either `href` field, `path`, `onClick`, or `items`, but not together. If I hadn't put the never, it would have accepted if I passed an item with multiple of these fields.

dlrice (approved):

PR #158, opened 2021-01-18
Optional data loader props

dlrice:
## Purpose
In the UniProt website there are a few places where we use DataList, DataTable without needing the data-loading functionality (eg BlastResultTable). Doing this however throws some warnings about not passing the required onLoadMoreItems prop. This PR makes these props optional.

## Approach
Make onLoadMoreItems and hasMoreData optional and don't use the interaction observer if not needed.

## Testing
No tests updated.

## Stories
None updated.

## Checklist
- [x] For the stories you created/updated, are the props modifiable through knobs?

### DIFF
--- a/src/components/data-loader.tsx
+++ b/src/components/data-loader.tsx
@@ -5,6 +5,7 @@ import {
   useMemo,
   useCallback,
   ReactNode,
+  FC,
 } from 'react';
 
 import Button from './button';
@@ -42,9 +43,7 @@ type WrapperProps = {
   clickToLoad?: boolean;
 };
 
-const withDataLoader = (
-  BaseComponent: FC<{ data: unknown[] }>
-) => {
+const withDataLoader = (BaseComponent: FC<{ data: unknown[] }>) => {
   const Wrapper: FC<WrapperProps> = ({
     onLoadMoreItems,
     hasMoreData,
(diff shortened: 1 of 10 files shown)


dlrice:
Exported DataTable and DataTableWithLoader (and likewise for cards). Also changed the stories to make it clear these two options exist:
Screenshot 2021-01-18 at 14 20 18

src/components/data-loader.tsx
```diff
@@ -8,30 +14,65 @@ import '../styles/components/data-loader.scss';
...
+  data: unknown[];
+  /**
+   * Use a button to load more data instead of having infinite scrolling
+   */
+  // eslint-disable-next-line react/require-default-props
+  clickToLoad?: boolean;
+};
+
+const withDataLoader = (
+  BaseComponent: (props: { data: unknown[] }) => JSX.Element
```
aurel-l, line 46: Would suggest using 
```suggestion
  BaseComponent: FC<{ data: unknown[] }>
```

src/components/data-loader.tsx
```diff
@@ -8,30 +14,65 @@ import '../styles/components/data-loader.scss';
...
+) => {
+  const Wrapper = ({
```
aurel-l, line 48: And 
```suggestion
  const Wrapper: FC<WrapperProps> = ({
```
As FC types children for us, and in a slightly more generic way using `ReactNode`rather than using `JSX.Element`

aurel-l (approved):

PR #159, opened 2021-01-18
Intersection observer mocking

dlrice:
## Purpose
Test coverage decreased in last PR so needed to bring it back up. Noticed in the cov report the ioSupport wasn't tested so this PR tests this.

## Approach
Mocking IntersectionObserver. Had to push `ioSupport` test inside of `withDataLoader`.

## Testing
`should show click-to-load if no IntersectionObserver support`

## Stories
N/A

## Checklist
 also: add key in story DecoratedListItem

### DIFF
--- a/src/components/data-loader.tsx
+++ b/src/components/data-loader.tsx
@@ -13,8 +13,6 @@ import Loader from './loader';
 
 import '../styles/components/data-loader.scss';
 
-const ioSupport = 'IntersectionObserver' in window;
-
 type WrapperProps = {
   /**
    * Callback to request more items if user scrolled to the bottom of the scroll-container or if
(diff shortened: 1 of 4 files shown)


src/components/data-loader.tsx
```diff
@@ -82,6 +80,8 @@ const withDataLoader = (BaseComponent: FC<{ data: unknown[] }>) => {
       handleAskForMoreData();
     };
 
+    const { current: ioSupport } = useRef('IntersectionObserver' in window);
```
aurel-l, line 83: If feel this got moved here because of limitations in mocking in the tests.
I'd say in this case just replace both uses of `ioSupport` with the actual `'IntersectionObserver' in window` statement.

dlrice: Updated.

aurel-l (approved):

PR #160, opened 2021-01-20
separate Facet from Facets components

aurel-l:
## Purpose
Expose the `Facets` inner logic into an other components.

## Approach
Extract a standalone `Facet` components, aimed to be used within a `Facets` component.
`Facets` can be used as before, but any child will be added as an other item within it.

## Testing
Existing tests still passing,
Added tests for `Facet` and `Facets`

## Stories
Story modified to present new usage

## Checklist


src/components/facets.tsx
```diff
@@ -81,96 +81,135 @@ export const stringify = (
...
 }) => {
   const location = useLocation();
+  const search = parse(location.search, queryStringKey);
 
-  if (!(data && data.length)) {
+  if (!data.values?.length) {
     return null;
   }
+  return (
+    <div {...props}>
```
xwatkins, line 112: What extra props are expected here? Is it just `className`?

xwatkins (approved):
Just a comment about spreading `props` into a `<div>`, all good apart from that

src/components/facets.tsx
aurel-l: That could be anything really, including `className`
It's just to avoid wrapping this component in an unnecessary div if we ever need to add an `aria-` or `data-` attribute or anything else really.
We already started having this pattern accross new and modified Franklin components

PR #161, opened 2021-01-28
Switch to TypeScript & improve DataTable and DataList

aurel-l:
## Purpose
Switch to TypeScript & improve DataTable and DataList

## Approach
Switch to TypeScript, rewrite bunch of code

## Testing
Updated snapshots, stories, and tests

## Stories
All the ones changed in this PR

## Checklist
  - nope...
  - Yes, and also, updated old ones that should have been switched to TypeScript before (any `.tsx` component now has their related story and tests switched to TS too)

What I did, in order, has been:
 - Started as wanting to improve typing in DataTable to have the data type passed everywhere it was needed (if data is of type `D`, every other prop related to data handling like `getIdKey` would correctly be typed too)
 - Got lost in TypeScript generics definitions that infected more files...
 - **API change (non-breaking):** Improved the function props of `DataTable` and `DataList` to have a signature similar to array method (`(item, index, array) => ...`)
 - **API change (breaking):** Renamed `DataTable`'s `onSelect` to `onSelectRow` to avoid mixing it up with the standard `onSelect` prop available on all `HTMLElement`s as it might cause mix ups since the signature and behaviour is different.
 - **API change (non-breaking):** Removed `selectable` prop, as that can be deduced from the presence of not of the `selected` prop
 - **API change (non-breaking):** Make `onSelectRow` mandatory _only_ alongside a `selected` prop, and invalid without a `selected` prop.
 - **API change (non-breaking):** Make `sorted` field in a column only available if the column is marked as `sortable`, and made `label` optional.
 - **API change (breaking):** replace named exports for `density` with pure TypeScript enforcing of possible values.
 - **API change (breaking):** `getIdKey` is now mandatory, but strongly typed and has more arguments passed to it in order to create a useful id (`(datum: T, index: number, data: T[]) => string | number`)
 - Reviewed all the code that was supposed to have been switched to TS earlier and switched it.
 - Made sure to have TypeScript type checking applied to stories and tests too, and linting applied to TS stories. -> This caught quite a lot of unused or invalid code.


src/components/data-table.tsx
```diff
@@ -0,0 +1,239 @@
...
+  columns: Array<SortableColumn<T> | NonSortableColumn<T>>;
+  /**
+   * Table fixed layout
+   */
+  // Note sure why it doesn't detect it as used...
+  // eslint-disable-next-line react/no-unused-prop-types
+  fixedLayout?: boolean;
+};
+
+const BLOCK = 'data-table';
```
aurel-l, line 42: Tried this approach here of putting the BEM's `Block` bit inside a variable reused throughout the file.
Not really necessary, but I thought it might be easier to follow BEM concepts that way, and would avoid typos in class names as those can't be caught by tools.

src/styles/_mixins.scss
```diff
@@ -5,16 +5,6 @@
 $global-left: if($global-text-direction == rtl, right, left);
 $global-right: if($global-text-direction == rtl, left, right);
 
-@mixin chevron($direction, $color) {
```
xwatkins, line 8: I know we don't use it anywhere else at the moment but it might be a better way to draw chevrons than to use the SVGs as we're doing now in some components? Not saying to do all the changes now, but to keep the chevron generation as a mixin?

xwatkins (approved):

PR #162, opened 2021-02-04
add a visual cue that a column is sortable even though it is not sorted

aurel-l:
## Purpose
Add a third state to a column header where it can be visually marked as sortable even though it is not sorted.

## Approach
Visual change, opacity 0.7 on a sortable column

## Testing
Updated snapshots, checked story

## Stories
data table story

## Checklist

### DIFF
--- a/src/components/__tests__/__snapshots__/data-table.spec.tsx.snap
+++ b/src/components/__tests__/__snapshots__/data-table.spec.tsx.snap
@@ -13,15 +13,17 @@ exports[`DataTable should render autoload 1`] = `
            
         </th>
         <th
-          class="data-table__header-cell--ascend"
+          class="data-table__header-cell--sortable data-table__header-cell--ascend"
         >
           Column 1
         </th>
-        <th>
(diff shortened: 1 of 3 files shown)


xwatkins (commented):
Not sure whether this would warrant a separate PR but I think the whole header should be the trigger?

aurel-l:
> Not sure whether this would warrant a separate PR but I think the whole header should be the trigger?

Be the trigger of what? And do you mean the whoooole table header (as in, the whole set of column headers?)?

xwatkins:
Actually it is already, it's just that I think `cursor: pointer;` should be on `&--sortable` not just the `:after`

aurel-l:
> Actually it is already, it's just that I think `cursor: pointer;` should be on `&--sortable` not just the `:after`

Ah, I see. I wasn't sure what you were refering to as the behaviour was there already. I have now modified the cursor style to cover the whole column header.

xwatkins (approved):

PR #165, opened 2021-02-09
reduce markup complexity for cards

aurel-l:
## Purpose
Reduce markup complexity of the Card component, and use only one `section` element for each card

## Approach
Change markup

## Testing
Updated snapshots

## Stories
Card

## Checklist

### DIFF
--- a/src/components/__tests__/__snapshots__/card.spec.tsx.snap
+++ b/src/components/__tests__/__snapshots__/card.spec.tsx.snap
@@ -2,10 +2,10 @@
 
 exports[`Card component should render 1`] = `
 <DocumentFragment>
-  <div
+  <section
     class="card"
   >
-    <section>
+    <div>
       <div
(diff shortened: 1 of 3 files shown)


xwatkins (commented):
It's an improvement and more semantic, only thing is that the link borders now don't touch each other (but there was the issue of the the fact the "separator" was not in-between.
I actually wonder whether a separator is really needed, and the right amount of padding could work (see tabs for instance)

aurel-l:
> It's an improvement and more semantic, only thing is that the link borders now don't touch each other (but there was the issue of the the fact the "separator" was not in-between.
> I actually wonder whether a separator is really needed, and the right amount of padding could work (see tabs for instance)

Revert to previous way of handling separators, but without the need for an extra class for this

xwatkins (approved):

PR #167, opened 2021-02-24
loader component with progress bar

aurel-l:
## Purpose
Add a progress bar to the loader component

## Approach
Add a progress bar below the existing loading visualisation, which only appears later in order to not be visible if the loading time was short enough

## Testing
Updated story, created 2 unit tests

## Stories
Loader story, with added knob to change the progress value

## Checklist


dlrice:
Is the new background for loader intentional?
Screenshot 2021-02-25 at 11 04 20

aurel-l:
> Is the new background for loader intentional?
> Screenshot 2021-02-25 at 11 04 20

This is just in the story as to make clear where the loader positions itself within its container. The loader component does not have a background set

xwatkins:
What is the purpose of the loading bar? I think you mentioned something to do with BLAST progress, but is there an accurate indication of how long it will take before it's ready? 
The use of loading bars has decreased over the years because they are generally misleading (how hasn't had an installation stuck at 98% for 20min!) and replaced with spinners (infinity, but something is happening) and changing information messages "retrieving data", "preparing alignment" etc...

aurel-l:
> What is the purpose of the loading bar? I think you mentioned something to do with BLAST progress, but is there an accurate indication of how long it will take before it's ready?
> The use of loading bars has decreased over the years because they are generally misleading (how hasn't had an installation stuck at 98% for 20min!) and replaced with spinners (infinity, but something is happening) and changing information messages "retrieving data", "preparing alignment" etc...

The purpose for it is definitely not to guess how much time is left for things we don't have control with.
If you look at the companion PR I'm just using it for download progress.
If I'm to retrieve 1.5G of results data, as a user I'd like to know at what pace the data I'm interested in is being downloaded, as to not rage-quit after 19 minutes when we could have shown that it was just missing 1% of download.

xwatkins:
> > What is the purpose of the loading bar? I think you mentioned something to do with BLAST progress, but is there an accurate indication of how long it will take before it's ready?
> > The use of loading bars has decreased over the years because they are generally misleading (how hasn't had an installation stuck at 98% for 20min!) and replaced with spinners (infinity, but something is happening) and changing information messages "retrieving data", "preparing alignment" etc...
> 
> The purpose for it is definitely not to guess how much time is left for things we don't have control over.
> If you look at the companion PR I'm just using it for download progress.
> If I'm to retrieve 1.5G of results data, as a user I'd like to know at what pace the data I'm interested in is being downloaded, as to not rage-quit after 19 minutes when we could have shown that it was just missing 1% of download.

In the context of download it makes sense as it is quantifiable. A lot of times these are implemented by checkpoints and guesstimating what time each process will take which is where it gets misleading! We should be careful where and when we use it.

xwatkins (approved):

PR #171, opened 2021-03-02
Add zoom icons

xwatkins:
Add new icons. SVGs were simplified through https://jakearchibald.github.io/svgomg/


src/svg/zoom-in.svg
```diff
@@ -0,0 +1 @@
+<svg viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M7.5 3v2.5H10v2H7.5v2.499h-2v-2.5H3v-2h2.5V3h2z" fill="#000"/><path fill-rule="evenodd" clip-rule="evenodd" d="M15.78 13.833l-3.116-3.115a.75.75 0 00-.53-.219h-.51a6.468 6.468 0 001.375-4c0-3.59-2.91-6.499-6.5-6.499S0 2.91 0 6.5s2.91 6.499 6.5 6.499c1.509 0 2.896-.513 3.999-1.375v.51c0 .2.078.39.219.53l3.115 3.116a.747.747 0 001.06 0l.884-.885a.753.753 0 00.003-1.062zm-4.281-7.334a5 5 0 11-10 0 5 5 0 0110 0z" fill="currentColor"/></svg>
```
aurel-l: There's an issue here where the "+" sign won't change colour with the rest of the icon because of the hardcoded `fill="#000"`.
I usually don't set the fill attribute within the SVG (unless need to override), and just set it to `fill="currentColor"` on the SVG element itself. All the elements below will inherit from it. This way I save in bytes by just defining it in one place, then I avoid this kind of issue where one might worget to change an instance of fill within the SVG itself.

xwatkins: Sorry, that was an oversight, I must have pushed the update while you were writing this...

aurel-l: Also, for readability, I'd be against removing new lines and indentations, it makes it overly difficult to debug, and I believe white spaces in a text file could be considered free when gzipping, and can also be easily optimised by a tool at compile time (unlike the precision of the values which require a human eye to check if the image is still visually similar, which is why the mentioned svgomg tool is still important)

aurel-l (changes requested):
Changes to the 3 SVGs as per comment

aurel-l:
Added a way to change the colour of the icons within the corresponding story

xwatkins:
> Added a way to change the colour of the icons within the corresponding story

Nice one. Waiting for the updated icon from @AustraAnna then will look at indentation. It came out of svgomg like that, so I might have to look at the settings.

xwatkins:
Updated icons

aurel-l (approved):

PR #173, opened 2021-03-04
External link with fallback to tidy url

dlrice:
## Purpose
If no link text is passed to an external link then fallback to using the url.

## Approach
* Copied @aurel-l's regexp from the website and put it in a function called `tidyUrlString`
* Conditioned on existence of the `children` prop in the ExternalLink to determine what to show
* As I touched `utils.jsx`, `utils.spec.jsx`,  `window.jsx`, `window-buttons.jsx` I converted to tsx.
* Replaced `getClassNames` with `classnames` to avoid adding types to the fn.
* `window.spec.jsx.snap` has been updated due to an unrealised bug

## Testing
Unit tests and snapshots

## Stories
Added an ExternalLink story without passing link text

## Checklist


src/components/long-number.tsx
```diff
@@ -9,7 +9,8 @@ type Props = {
   children: string | number;
 };
 
-const LongNumber: FC<Props> = ({ children }: Props) =>
-  formatLargeNumber(children);
+const LongNumber: FC<Props> = ({ children }) => (
+  <>{formatLargeNumber(children)}</>
```
dlrice, line 13: Had to use a fragment to make TSC happy

aurel-l: In that case were you are specifying the type of `children` and explicitely returning a Fragment you might be able to leave out the FC annotation and just type this as a standard function (`const LongNumber = ({ children }: Props) => (`)

src/components/window/window.tsx
```diff
@@ -0,0 +1,74 @@
...
+    width,
+    minHeight: height,
+    top: `calc((80vh - ${height}) / 2)`,
+    left: `calc((100vw - ${width}) / 2)`,
+  };
+  const baseClassName = 'window';
+  const cssClasses = cn(baseClassName, className, {
+    [`${baseClassName}--with-shadow`]: !!withShadow,
+  });
+  if (typeof onWindowOpen === 'function') {
```
aurel-l, line 46: Could be replaced with `onWindowOpen?.()`

src/utils.tsx
```diff
@@ -59,25 +73,7 @@ export function highlightSubstring(string, substring) {
...
-    return mixedInput
-      .map((item) => getClassNames(item, optionalCondition, output))
-      .filter((x) => x) // removes nulls, undefineds, empty strings, etc.
-      .join(' ');
-  }
-
-  return output;
-}
+export const tidyUrlString = (url: string) => {
+  const re = /(^(https?:\/\/)?(www\.)?)|(\/+$)/g;
```
aurel-l, line 77: 2 things:
- Maybe put it outside the function, so that it's compiled once and reused.
- Should URLs like `//:www.ebi.ac.uk/` be cleaned too?

dlrice: - Yep, first point incorporated.
- `//:www.ebi.ac.uk/` are you thinking to handle the `//:` at the start or if we should be leaving `www` there?

src/components/window/window.tsx
dlrice: Updated

src/components/long-number.tsx
dlrice: Updated

src/utils.tsx
dlrice: I'm thinking it would be useful to use https://www.npmjs.com/package/valid-url here to complain in the console so it will be picked up by our error logging service that we have set up

aurel-l: I was think `//:www.ebi.ac.uk/` should also become `ebi.ac.uk`, not just for `https//:www.ebi.ac.uk/`

aurel-l: I'm not too keen on adding a library for something that will be just used for content that we generate and have control over. Also, I'm more worried about dead links than invalid links in the context of the links we are generating, and this wouldn't catch it

aurel-l (approved):
For me it looks OK. Just make sure that it's OK with @xwatkins too to have the URL cleaning on by default

src/utils.tsx
dlrice: * `//:www.ebi.ac.uk/` isn't a valid url (at least it won't open anything in the browser) so the link itself would be broken so not sure if we need to try to handle formatting invalid urls. I am currently handling multiple trailing slashes just because that seems to work in browsers eg `https://wwwdev.ebi.ac.uk////`.
* Yeah deadlinks are a concern but I would also imagine an invalid external link that we provide on the site would want to be detected and fixed if say the url from the API has a typo. As we have so many external links throughout the site my thought is this would be an easy way to detect them as users view them on a page. This can point can be reassessed in the future post alpha.

aurel-l: Sorry, I realise I mixed up the URL at the beginning and copy pasted afterwards without realising. I meant `//www.ebi.ac.uk/`

dlrice: This is now handled

aurel-l (approved):

aurel-l (approved):

PR #176, opened 2021-03-15
Header secondary nav

xwatkins:
## Purpose
Add a secondary navigation section in the header component. This will be used on the website to link to help, contact. Also add a `subtext` which will be used for the release and statistics link.

## Approach
Add optional props

## Testing
Added to the render test

## Stories
Added example to storybook stories

## Checklist
- [] For the stories you created/updated, are the props modifiables through knobs?

### DIFF
--- a/src/components/header.tsx
+++ b/src/components/header.tsx
@@ -64,6 +64,33 @@ const HeaderItem: FC<HeaderItemProps> = ({ item }) => {
   );
 };
 
+const HeaderListItem: FC<{
+  item: HeaderPossibleItem | HeaderDropdown;
+  isNegative: boolean;
+}> = ({ item, isNegative }) => (
+  <li>
+    {item.items ? (
+      <DropdownButton
+        label={item.label}
+        className={cn({
+          'dropdown-container__trigger--negative': isNegative,
+        })}
+        openOnHover
+      >
+        <ul>
+          {item.items.map((subItem, index) => (
+            <li key={typeof subItem.label === 'string' ? subItem.label : index}>
+              <HeaderItem item={subItem} />
+            </li>
+          ))}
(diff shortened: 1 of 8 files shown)


src/components/header.tsx
```diff
@@ -77,46 +104,59 @@ type HeaderProps = {
...
    */
   search?: ReactNode;
+  /**
+   * Secondary items
+   */
+  secondaryItems?: Array<HeaderPossibleItem | HeaderDropdown>;
+  /**
+   * Subtext
+   */
+  subtext?: ReactNode;
```
xwatkins, line 114: Actually realised there is an issue if providing `subtext` and no `secondaryItems`. Will update tomorrow

xwatkins: Done

aurel-l:
Would be good to reduce the number of stories and have the negative flag and the fact that it has a search area, secondary items, and a subtext toggable through boolean knobs in order to visually check all the combinations

xwatkins:
> Would be good to reduce the number of stories and have the negative flag and the fact that it has a search area, secondary items, and a subtext, all togglable through boolean knobs in order to visually check all the combinations

Good idea, nice way to check that all combinations are working

aurel-l:
OK, so now that I can combine things more easily I see that the subText, when used with the search area and the right icons, is pushing the header to be taller, and also close, but not completely aligned with the bottom of the search area.
As the top of the icons on the right aligns with the top of the search bar, the bottom of the subtext should probably align with the bottom of the search area.

aurel-l (approved):
I assume the text wrapping is on purpose, otherwise you need a `white-space: nowrap` on the small tag for the subtext.

PR #177, opened 2021-03-17
Fix issue with column size not being applied

xwatkins:
## Purpose
Column sizing was being overwritten in css

## Approach
Move css block to be more specific.

## Testing
Visual check in stories

## Checklist

### DIFF
--- a/src/styles/components/data-table.scss
+++ b/src/styles/components/data-table.scss
@@ -31,59 +31,58 @@
     top: 0;
     z-index: $z-index-low;
     position: sticky;
-  }
 
-  &__header-cell {
-    &--sortable {
-      cursor: pointer;
+    &.data-table__header-cell {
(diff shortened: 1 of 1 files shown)


aurel-l (approved):

aurel-l:
I understand this is for the table with the `--checkbox` class?
No visual change in the story so I'm assuming that at least it still does what it's supposed to do regarding the sortable column headers

xwatkins:
> I understand this is for the table with the `--checkbox` class?
> No visual change in the story so I'm assuming that at least it still does what it's supposed to do regarding the sortable column headers

You can actually see it in the story where the table has the "fixed" option. Before the fix the checkbox column was taking up all the space available because the width was overwritten.

aurel-l (approved):

PR #178, opened 2021-03-25
changes to tile and more stuff

aurel-l:
## Purpose
Changes to the Tile component and other related changes needed for the Home page.

## Approach
same as purpose

## Testing
Update snapshot, update stories

## Stories
Tile story -> less stories, now with more knobs

## Checklist

What changed:
 - Tile component API
   - props changes
   - now renders actual links
 - Accessibility changes, separate what headers are rendered (configurable), but the style is defined and independent to the chosen heading level.
 - Update uniprot reviewed and unreviewed icons
 - Add .visually-hidden helper class to define markup useful only for semantic structures of the page and/or for screen reader navigation
 - Add text size helper functions, meant to be used instead of specific `h<x>` for text sizing.

Dependency of https://github.com/ebi-uniprot/uniprot-website/pull/326


src/components/tile.tsx
```diff
@@ -35,42 +40,73 @@ type Props = ButtonHTMLAttributes<HTMLButtonElement> & {
...
-    <section className="tile__background-image">{backgroundImage}</section>
-    <section className="tile__content">
-      <h3 className="tile__header">{title}</h3>
-      {subtitle && <h5 className="tile__subtitle">{subtitle}</h5>}
-      {description && (
-        <small className="tile__description">{description}</small>
+}) => {
+  const isExternal = 'url' in props;
+
+  const linkTarget = (
```
xwatkins, line 65: Wonder if "tileBody" or "tileMain" would be a better name?

src/styles/components/tile.scss
```diff
@@ -31,12 +35,32 @@ $tile: '.tile';
     margin: 0;
     line-height: 1.2;
   }
-  padding: $global-padding/2 0 0 0;
-  cursor: pointer;
 
   &__header {
+    @include fs-font-settings('H3');
     display: block;
+    margin-top: 0.5em;
```
xwatkins, line 41: should this be `rem`?

src/styles/components/tile.scss
```diff
@@ -31,12 +35,32 @@ $tile: '.tile';
...
+
+    // match mixin specificity...
+    @media (min-width: 0px) {
```
xwatkins, line 44: What does this do?

src/styles/components/tile.scss
```diff
@@ -31,12 +35,32 @@ $tile: '.tile';
     margin: 0;
```
xwatkins, line 35: Not sure what's caused this change, but the spacing between header and subheader is too narrow now. Eyeballing it at `$global-margin/2`

xwatkins (changes requested):

src/components/hero-container.tsx
```diff
@@ -1,21 +1,19 @@
+import { createElement, FC, HTMLAttributes } from 'react';
 import cn from 'classnames';
-import { FC, ReactNode } from 'react';
 
 import '../styles/components/hero-container.scss';
 
-type HeroContainerProps = {
+type HeadingLevels = `h${1 | 2 | 3 | 4 | 5 | 6}`;
```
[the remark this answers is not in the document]
aurel-l: I don't think we have a common TS file, should we create one? For typography what we have is a scss partial file

src/styles/components/tile.scss
aurel-l: You tell me, I'm not attached to anything. rem will be fixed, em will give us more or less space depending on the font-size of the title (in case it's overriden)

src/styles/components/tile.scss
aurel-l: It just matched the specificity of `fs-font-settings` when I was using it, but with the new helper classes it's not needed anymore

src/styles/components/tile.scss
aurel-l: Actually, not needed now using the new helper classes. I think the spacing is now correct

src/styles/components/tile.scss
aurel-l: I think it now has correct spacing

src/components/tile.tsx
aurel-l: changed

src/components/hero-container.tsx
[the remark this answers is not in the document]
aurel-l: created

xwatkins (approved):

PR #185, opened 2021-04-16
handle stale data in data table and data list

aurel-l:
## Purpose
Prevent crashing if we know the data is loading

## Approach
Optionally pass an additional loading flag to DataTable or DataList, if a renderer crashes while this flag is true, don't render anything, just a loader if it's the first cell.

## Testing
Story

## Stories
DataTable & DataList

## Checklist

### DIFF
--- a/src/components/data-table.tsx
+++ b/src/components/data-table.tsx
@@ -2,6 +2,7 @@
 import { memo, useCallback, ReactNode, HTMLAttributes, FormEvent } from 'react';
 import cn from 'classnames';
 
+import Loader from './loader';
 import withDataLoader, { WrapperProps } from './data-loader';
 
 import '../styles/components/data-table.scss';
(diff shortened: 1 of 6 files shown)


src/components/data-table.tsx
```diff
@@ -85,30 +86,50 @@ const DataTableHead = <Datum extends BasicDatum>({
...
+  } catch (error) {
+    /**
+     * We get here only if the renderer fails. If the renderer returns null of
+     * undefined because of a lack of data, then it will no throw and will not
+     * display the loader at all
+     */
+    if (!loading) {
+      throw error;
+    } else {
+      rendered = firstColumn && <Loader />;
```
xwatkins, line 116: I'm not sure about this approach. There will be cases where a few columns might be invalid and it doesn't look great. Maybe we could check with UX if there could be a less "in your face" loading indicator we could use here. In the meantime we can keep this as a reminder something should be done about it, or return null, not fussed.

image

xwatkins (approved):

src/components/data-table.tsx
aurel-l: Should I propose an alternative visual now, or do you want to get that published asap to be able to use it in your current changes?

xwatkins: Let's have that discussion on #ux but you can publish now. We currently have 2 different levels of loading (Loader and the spinner) but it's not really well defined on Franklin what should be used there so this could be done at the same time as adding a column/table (card too?) loading indicator.

xwatkins: I've got it linked locally BTW so not in an immediate rush

aurel-l: published

PR #187, opened 2021-04-27
refactor cards

aurel-l:
## Purpose
Refactor Card component

## Approach
- Accept ReactNode as text for the links at the bottom of a card
- Replace title and subtitle with generic header prop (responsibility of the user of the component to use the correct heading level)
- Made header separator optional (through props)
- Remove the onClick in favour of using a proper link in the back
- Let the clicks go through to the new background link, except for `<a>`, `<button>`, and `<input>`.
- Removed all but one story, use knobs

## Testing
Updated tests and snapshots

## Stories
Card

## Checklist
  - Reduced codebase

Doubts: Is the `active` prop useful/used anywhere?

### DIFF
--- a/src/components/card.tsx
+++ b/src/components/card.tsx
@@ -43,11 +43,6 @@ type Props = {
    */
   links?: Array<CardActionProps>;
   /**
-   * Should the card styling show it as active or not
-   */
-  // TODO: where is that used?
-  active?: boolean;
-  /**
    * Target/link of the card when clicking on it
    */
   to?: LinkProps['to'];
@@ -61,7 +56,6 @@ const Card = forwardRef<HTMLElement, Props & HTMLAttributes<HTMLElement>>(
       children,
       links,
       to,
-      active,
       className,
       ...props
     },
@@ -69,7 +63,6 @@ const Card = forwardRef<HTMLElement, Props & HTMLAttributes<HTMLElement>>(
   ) => (
     <section
       className={cn(className, 'card', {
-        'card--active': active,
         'card--has-link': to,
       })}
       ref={ref}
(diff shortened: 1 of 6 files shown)


src/components/card.tsx
```diff
@@ -23,72 +31,87 @@ const CardLink: FC<CardLinkProps> = ({ name, link, color }) => (
...
+  headerSeparator?: boolean;
   /**
    * Links to be displayed at the bottom of the card
    */
-  links?: Array<CardLinkProps>;
+  links?: Array<CardActionProps>;
   /**
    * Should the card styling show it as active or not
    */
+  // TODO: where is that used?
```
xwatkins, line 48: This should have been used whenever a Card checkbox is selected. It looks like it's been lost, but also I think we need to think about it carefully as we have:
- mouseover
- consistency with selection in the results table
- there was mention of alternate card background colour like in results table (not in favour of that)

aurel-l: I'll remove it then since the checkbox is not handled by the card

xwatkins: Mmm I don't know, I'd rather have checkbox as a modifier on the card pattern, as it is included in the data table pattern and we could then share styles between them for consistency. Something else I was noticing while using the website was that I think it would be good to increase the clickable area for the Card checkbox as it's easy to miss and then navigate to wherever the card links to.

aurel-l: Let's make a Jira for that for later then.
There's so much more to handle if we want to handle checkboxes in a correct way:
 - Accessibility
 - Target size
 - Who has the control of the checkbox state (uniprot-website or franklin's card?)
 - Ctrl/Cmd-click and Shift-click behaviours
 - ... and probably more

xwatkins: OK - I thought the Card used to have the checkbox as part of it but it looks like it didn't. Must have happened in another refactor!

xwatkins (approved):

src/components/card.tsx
aurel-l: It used to "assume" it contained a checkbox, but in its child, because there was style for that that I removed because it didn't make any difference in where we used it.

Issue #188, opened 2021-04-29
Evedence tag


AustraAnna:
**Name of component**
Evidence tag

**Component modifier**
evidence tag, mouse over and "pressed", "selected"

**Description of the change**
Colour overlapping with other design areas - regular - fill Gainsboro stroke Weldon Blue, mouse over - fill Pastel blue, stroke Weldon Blue, "pressed" - fill Independence 70 %, stroke Weldon Blue

**Add Screenshots / Mockups if relevant**

![Screenshot 2021-04-29 at 17 31 29](https://user-images.githubusercontent.com/72505072/116585900-c2573c80-a910-11eb-8cf3-42eb6038a55b.png)

https://www.figma.com/file/D2euYJHUDqsu4m1SqfWl6p/Help?node-id=808%3A4068

xwatkins:
Done

PR #189, opened 2021-04-30
Evidence tag changes

xwatkins:
## Purpose
New styling as defined in #188 

## Approach
Using pseudo-classes to define the new behaviour. I don't think `:active` is quite the behaviour defined by "when pressed" as it disappears after the click, but focus remains which I think is ok.
Move to TS (slight drop in coverage)

## Testing
Updated tests

## Stories
Added knobs to the story

## Checklist

### DIFF
--- a/src/components/evidence-tag.tsx
+++ b/src/components/evidence-tag.tsx
@@ -1,5 +1,12 @@
-import { useState, useRef, cloneElement } from 'react';
-import PropTypes from 'prop-types';
+import {
+  useState,
+  useRef,
+  cloneElement,
+  ReactNode,
+  FC,
+  ReactElement,
+} from 'react';
+import cn from 'classnames';
 import { v1 } from 'uuid';
 
 import EvidenceTagIcon from '../svg/evidence-tag.svg';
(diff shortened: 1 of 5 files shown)


src/components/evidence-tag.tsx
```diff
@@ -8,13 +15,42 @@ import '../styles/components/evidence-tag.scss';
...
+   */
+  className?: string;
+  /**
+   * Decides the colour of the tag
+   */
+  iconComponent: ReactElement<{ width: number; height: number }>;
+  /**
+   * The content of the tag
+   */
+  children: ReactNode;
```
aurel-l, line 38: If you use FC, this is not needed

src/components/evidence-tag.tsx
```diff
@@ -8,18 +15,45 @@ import '../styles/components/evidence-tag.scss';
 
 const size = 12;
 
-const EvidenceTag = ({ label, title, className, iconComponent, children }) => {
+type EvidenceTagProps = {
+  /**
+   * Displayed on the tag
+   */
+  label: string;
```
aurel-l, line 22:
```suggestion
  label: ReactNode;
```

xwatkins: I think we want to be quite restrictive about that and only allow strings.

src/components/evidence-tag.tsx
```diff
@@ -8,18 +15,45 @@ import '../styles/components/evidence-tag.scss';
...
+  /**
+   * Displayed on on mouseover
+   */
+  title: string;
```
aurel-l, line 26: I would remove that as it's now included in `HTMLAttributes<HTMLButtonElement>>`, and also just let it flow with the rest of the `props` to the button instead of the span within

src/components/evidence-tag.tsx
```diff
@@ -8,18 +15,45 @@ import '../styles/components/evidence-tag.scss';
...
+  className,
+  iconComponent = <EvidenceTagIcon />,
+  children,
+  ...props
+}) => {
   const idRef = useRef(v1());
   const [contentDisplay, setContentDisplay] = useState(false);
   return (
     <>
       <button
```
aurel-l, line 41: Is there enough overlap with the `Button` component to be able to use it instead and avoid `evidence-tag`-specific styling?

aurel-l (approved):

PR #196, opened 2021-05-12
2021 05 main search

xwatkins:
## Purpose
Add options for secondary buttons in the _Main Search_

## Approach
Add a prop which is an array of labels and actions.
Also converted some components to TS

## Testing
No extra tests, but could add some later.

## Stories
Added a story

## Checklist


src/components/main-search.tsx
```diff
@@ -0,0 +1,141 @@
...
+                </li>
+              ))}
+            </ul>
+          )}
+        </DropdownButton>
+      )}
+      <div className="main-search__input">
+        <input
+          type="text"
+          className="main-search__input"
```
aurel-l, line 102: Is the duplicated class name on 2 different elements on purpose?

aurel-l (approved):
Looks good 👍🏽 
I did have a go at extra changes (mainly to add more TypeScript), so please have a look if you're happy with those before merging

xwatkins:
Having a look. Also forgot to add the transparency change on hover for the secondary buttons. Made changes to `2021_06-header_changes` in the meantime

src/components/dropdown-button.tsx
```diff
@@ -121,11 +121,7 @@ const DropdownButton = ({
         style={style}
       >
         {showMenu &&
-          (childType === 'function'
-            ? (children as (
-                showMenu: Dispatch<SetStateAction<boolean>>
-              ) => ReactNode)(setShowMenu)
-            : children)}
+          (typeof children === 'function' ? children(setShowMenu) : children)}
```
xwatkins, line 124: Magic!

aurel-l: TypeScript magic 😅 
You were using a variable that had been set with the same statement, but TypeScript lost track of what that variable meant when you were using it here, so I just reused the statement here so that TypeScript could infer types correctly here

PR #198, opened 2021-05-20
Add new Ellipsis reveal pattern

xwatkins:
## Purpose
Need a new pattern to reveal extra text which is not in list form.

## Approach
Simple button to reveal  text.

## Testing
Test that clickling on the button reveals the hidden text.

## Stories
Simple story added

## Checklist


src/components/__tests__/ellipsis-reveal.spec.tsx
```diff
@@ -4,9 +4,12 @@ import EllipsisReveal from '../ellipsis-reveal';
...
     render(<EllipsisReveal>Some other text</EllipsisReveal>);
+
+    expect(screen.queryByText('Some other text')).not.toBeInTheDocument();
+
     const button = screen.getByRole('button');
     fireEvent.click(button);
-    const expandedText = screen.findByText('Some other text');
-    expect(expandedText).toBeTruthy();
+
+    expect(screen.getByText('Some other text')).toBeInTheDocument();
```
aurel-l, line 13: I changed the test, mainly because as `findBy...` returns a promise, it would always have been truthy

aurel-l:
Did some changes on the style (use CSS instead of inline style), test (see specific comment), and pass more props to the underlying Button component.
Feel free to re-review

aurel-l (approved):

aurel-l:
Does this need an extra prop, for the cases where it's used with more text than just `[...]`? E.g. `view abstract [...]`

xwatkins:
> Does this need an extra prop, for the cases where it's used with more text than just `[...]`? E.g. `view abstract [...]`

Maybe... I don't really think like "View abstract" is part of the same pattern though. My preference would be to have "View abstract" as a button like we have "Show Sequence" / "Hide Sequence" for the representative in UniRef https://beta.uniprot.org/uniref/UniRef100_A0A0B8QSR6 - it would have to be a 3rd level button though, as issues were raised when it was level 2 that it wasn't compact enough ;-)

xwatkins:
> Did some changes on the style (use CSS instead of inline style), test (see specific comment), and pass more props to the underlying Button component.
> Feel free to re-review

I was being lazy re: inline style as it was only 1 attribute. And good catch on the truthy promise...

aurel-l:
> I was being lazy re: inline style as it was only 1 attribute

Ah, I assumed it was to override CSS specificity, as just with one class it was not enough. But I did it with 2 classes, to match specificity and this way it's still overridable in uniprot-webiste if we really need to without the need for `!important`

PR #201, opened 2021-06-04
Autocomplete typescript

dlrice:
## Purpose
Needed as part of [TreeSelect with autocomplete closes when making selection task](https://www.ebi.ac.uk/panda/jira/browse/TRM-25862)

## Approach
`Autocomplete` and `AutocompleteItem` converted to functional components in typescript.

## Testing
Updated unit tests and snapshots.

## Stories
Story renamed as `.ts`

## Checklist


src/components/__tests__/__snapshots__/autocomplete.spec.tsx.snap
```diff
@@ -30,7 +30,9 @@ exports[`Autocomplete component should render 1`] = `
     >
       <div
         class="dropdown-menu__panel"
-      />
+      >
+        <ul />
+      </div>
```
aurel-l, line 35: Probably shouldn't render an empty list. Not sure if it's just because the data used for this test is not realistic and this won't happen in the real world

src/components/autocomplete.tsx
```diff
@@ -0,0 +1,181 @@
...
+  onChange,
+  showDropdownUpdated,
+  clearOnSelect = false,
+  placeholder = '',
+  filter = true,
+  value = '',
+  minCharsToShowDropdown = 0,
+  isLoading = false,
+  autoFocus = false,
+}: AutocompleteProps) => {
```
aurel-l, line 59: Would probably be good to have `...props` and pass it down either to the underlying `div`, or maybe better to the `SearchInput` component

aurel-l: done

src/components/__tests__/__snapshots__/autocomplete.spec.tsx.snap
aurel-l: changed

aurel-l (approved):

aurel-l (approved):
Applied my own suggestions to the PR as I figured you'd be busy with other things at the moment. If something's wrong feel free to revert

PR #204, opened 2021-06-16
Allow tree-select item type to be extendable (while supporting a basic item type).

dlrice:
## Purpose
Types need updating to work with uniprot website

## Approach
Make tree-select item extendable with generics

## Testing
N/A

## Stories
N/A

## Checklist


aurel-l (approved):
All good.

Not too comfortable with the generics and the recursivity in `getFlattenedPaths` and I'm not sure about the quality of my review but I guess as long as it works and it's tested then all good 👍🏽

PR #207, opened 2021-06-22
2021 08 tree select bugs

dlrice:
## Purpose
A few bugs seen with the TreeSelect:

1.  In Safari and Firefox (couldn't reproduce myself with FF89), when clicking on a TreeSelect parent element, the containing dropdown menu closes instead of showing that element's children ([jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-25862))
2. In Safari, when selecting an autocomplete item the click isn't registered
3. In all browsers, the tree node arrow sometimes appears on the next line

## Approach

1. 1 & 2: Comment out `onBlur` at line 101 of dropdown-button. You can see the difference in behaviour between Chrome and Safari with the following code:
```
      onBlur={(e) => {
        console.log(
          e.currentTarget,
          e.relatedTarget,
          e.currentTarget.contains(e.relatedTarget as Node)
        );
      }}
```
As the comment states there is [an open bug report](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus) so not sure how else to deal with it at this present time? I can't see the effect of commenting out `onBlur` aside from eliminating the observed bugs.

2. CSS updates

## Testing
None

## Stories
I added `onClick` actions to the button in the process of my debugging but thought might as well leave these in.

## Checklist

### DIFF
--- a/src/components/dropdown-button.tsx
+++ b/src/components/dropdown-button.tsx
@@ -98,9 +98,10 @@ const DropdownButton = ({
   return (
     <div
       className="dropdown-container"
-      onBlur={(e) =>
-        setShowMenu(e.currentTarget.contains(e.relatedTarget as Node))
-      }
+      // TODO: uncomment when Safari resolves bug: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus
+      // onBlur={(e) =>
+      //   setShowMenu(e.currentTarget.contains(e.relatedTarget as Node))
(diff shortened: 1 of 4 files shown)


aurel-l (approved):
Accepted, still a bit clunky on Safari (probably related to the linked bug report), especially regarding focus management, but at least it improved and fixed the existing bugs that were happening when using mouse and keyboard.

PR #217, opened 2021-08-10
Trm 26304 chip remove bug

dlrice:
## Purpose

- [Jira: BLAST: Can't remove organism chip, "x" not working](https://www.ebi.ac.uk/panda/jira/browse/TRM-26304)
- Adapt chip as Firefox won't fire onClick events on items within a button.

## Approach
- Convert chip from `<button>` to `<span>`

## Testing
- Updated unit tests and snapshots.
- Had to adjust coverage for some reason 🤷 

## Stories
- Added `onKeyPress` story

## Checklist

### DIFF
--- a/src/components/chip.tsx
+++ b/src/components/chip.tsx
@@ -79,7 +79,6 @@ export const Chip: FC<ChipProps & HTMLAttributes<HTMLButtonElement>> = ({
       )}
       onKeyPress={onKeyPress}
       onClick={onClick}
-      title={title}
       tabIndex={0}
       {...props}
     >
--- a/src/components/chip.tsx
+++ b/src/components/chip.tsx
@@ -41,6 +41,10 @@ type ChipProps = {
    * click event listener on the component (except on the close button if present)
    */
   onClick?: () => void;
+  /**
+   * key press event listener on the component
+   */
+  onKeyPress?: () => void;
 };
 
 export const Chip: FC<ChipProps & HTMLAttributes<HTMLButtonElement>> = ({
@@ -51,6 +55,7 @@ export const Chip: FC<ChipProps & HTMLAttributes<HTMLButtonElement>> = ({
   compact = false,
   title,
   onClick,
+  onKeyPress,
   ...props
 }) => {
   const onRemoveRef = useRef(onRemove);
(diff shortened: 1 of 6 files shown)


src/components/chip.tsx
```diff
@@ -65,23 +70,24 @@ export const Chip: FC<ChipProps & HTMLAttributes<HTMLButtonElement>> = ({
...
         'chip',
         { 'chip--disabled': disabled, 'chip--compact': compact },
         className
       )}
-      title={title}
+      onKeyPress={onKeyPress}
       onClick={onClick}
-      disabled={typeof disabled === 'boolean' ? disabled : !onClick}
+      title={title}
+      tabIndex={0}
```
aurel-l, line 83: I would prefer if the role of button is only applied if onClick and/or onKeyPress is set. Same thing for tabIndex.
I might lean towards checking if onClick exists, and if so render a button like before, and if not render a plain span, this is in order to avoid having to use roles and tab indices.

src/components/chip.tsx
```diff
@@ -65,23 +70,24 @@ export const Chip: FC<ChipProps & HTMLAttributes<HTMLButtonElement>> = ({
...
       className={cn(
         'chip',
         { 'chip--disabled': disabled, 'chip--compact': compact },
         className
       )}
-      title={title}
+      onKeyPress={onKeyPress}
       onClick={onClick}
-      disabled={typeof disabled === 'boolean' ? disabled : !onClick}
+      title={title}
```
aurel-l, line 82: This line could probably be removed altogether, the title prop will be passed down through the destructured `props` variable a bit below

src/components/chip.tsx
dlrice: Can do - just wondering what is the drawback of using roles and tab indices?

aurel-l: Not really a direct drawback, it's more to follow the first rule of ARIA which is ~don't talk about ARIA~ don't use ARIA if you can avoid it https://www.w3.org/TR/using-aria/#rule1

src/components/chip.tsx
```diff
@@ -68,25 +70,34 @@ export const Chip: FC<ChipProps & HTMLAttributes<HTMLButtonElement>> = ({
...
   return (
-    <span
-      role="button"
+    <Element
       className={cn(
         'chip',
         { 'chip--disabled': disabled, 'chip--compact': compact },
         className
       )}
+      type={element === 'button' ? 'button' : undefined}
```
dlrice, line 91: Couldn't figure out a way to assign to props instead of inline as kept getting a type error.

src/components/chip.tsx
dlrice: Had a go at doing this. Let me know what you think and if OK I'll update the unit tests to get more coverage.

src/components/chip.tsx
aurel-l: I guess that's fine like this, typing an extra props object with part of the props, but not all of them, and changing conditionally, also proved complicated for me to before.

aurel-l (approved):
All good, and if there's no easy test to add to improve coverage here, feel free to decrease the threshold

PR #220, opened 2021-08-17
Sliding panel changes

aurel-l:
## Purpose
To handle the use case of closing it if the button used to open it is clicked again. https://www.ebi.ac.uk/panda/jira/browse/TRM-26450

## Approach
 - Wait one frame before closing it, in order to flush the current event listeners, because closing is the last thing we'll want  to do
 - Better focus management for accessibility: Move focus in the first focusable element within when opening the panel, and restore it to where it was previously when closing it.

## Testing
Updated tests, added a test to check the focus management (improved coverage)

## Checklist

### DIFF
--- a/src/components/__tests__/sliding-panel.spec.tsx
+++ b/src/components/__tests__/sliding-panel.spec.tsx
@@ -1,13 +1,13 @@
-import { fireEvent, screen } from '@testing-library/react';
+import { fireEvent, screen, waitFor } from '@testing-library/react';
+import { sleep } from 'timing-functions';
 
 import renderWithRouter from '../../testHelpers/renderWithRouter';
 
 import SlidingPanel from '../sliding-panel';
 
 describe('SlidingPanel component', () => {
-  const onClose = jest.fn();
-
-  beforeEach(() => {
+  it('should call onClose when mouse clicks outside but not inside', async () => {
+    const onClose = jest.fn();
     renderWithRouter(
       <>
         <div data-testid="outside-component" />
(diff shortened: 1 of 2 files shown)


src/components/sliding-panel.tsx
```diff
@@ -61,9 +65,51 @@ const SlidingPanel: FC<
 }) => {
   const node = useRef<HTMLDivElement>(null);
 
-  const onCloseRef = useRef(onClose);
+  const onCloseRef = useRef<(() => void) | null>(onClose);
```
dlrice, line 68: I've seen `useRef` for prop callbacks in a few places - I can't remember what the benefit of doing this is? Is it performance?

dlrice (approved):
Noticed that the previous focus that is restored after closing is the sliding window button (which makes sense) - is that the desired effect?

src/components/__tests__/sliding-panel.spec.tsx
```diff
@@ -16,18 +16,45 @@ describe('SlidingPanel component', () => {
         </SlidingPanel>
       </>
     );
+    fireEvent.click(screen.getByText('Sliding panel content'));
+    await waitFor(() => sleep(200));
```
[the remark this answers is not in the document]
aurel-l: Good point. I had previously tried to pass some kind of condition that would make sense in the waitFor, but since nothing changes at all (that's the whole point), I defaulted to just sleeping, but the waitFor is not really useful then 😅

src/components/sliding-panel.tsx
aurel-l: Because those are functions that the user of the component will define without necessarily having knowledge of the implementation of the component. Ideally it would be a memoized function that gets passed, but that doesn't have to be the case, it could be an inlined function for example.
So while in some other places we do that for performance, here it's actually critical because we have 2 useEffects that depend on the fact that they're run either once only, or only on path changes, and they both use that onClose function, so if for whatever reason onClose get re-created in uniprot-website (because of a rerender of an unrelated prop) it would trigger those useEffects incorrectly.

aurel-l: Regarding the suggested change, at the moment TS deduces the type from the component props, so I don't think that is necessary to specify it again here, and in any case, we don't have the case `| null` because onClose is required

aurel-l:
> is that the desired effect?

Yep it is, that is part of accessibility guidelines about these kind of patterns. It's usually well described for modals, so here you can find a reference: https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html Accessibility features > 2 > 1

src/components/sliding-panel.tsx
dlrice: We're making the assumption then that `onClose` won't change after mounting the component. Which makes sense - I'm not sure when a consumer would want to change the `onClose` callback when it's already open.

aurel-l: Yeah, I don't think it would happen either, but it is a caveat with this approach: "when did the user actually really wanted to change this?"

PR #221, opened 2021-08-18
TreeSelect

aurel-l:
## Purpose
TreeSelect refactor, including fixes for https://www.ebi.ac.uk/panda/jira/browse/TRM-25739 , https://www.ebi.ac.uk/panda/jira/browse/TRM-25740 and https://www.ebi.ac.uk/panda/jira/browse/TRM-26452

## Approach
 - Fixed arrow going to next line: changes in CSS
 - Also fixed/cleaned up more CSS that wasn't actually applied, added animation for arrow, change selectors to use aria-attributes instead of custom classes
 - Improve accessibility by following [this example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html) regarding aria attributes and roles
 - Applied change in logic in order to open the single children by default (regardless of nested level)
 - Fixed logic to toggle open/close of nodes
 - Fixed double scrollbar

## Testing
 - Added specific test to cover the logic to get all the nodes that need to be opened by default
 - Added more edge-case nodes to the test data used in storybook too, updated snapshots.

## Stories
 - Tree select: newer mock data includes a node with a long text, containing a single child node with more children to open by default.

## Checklist

### DIFF
--- a/src/components/tree-select.tsx
+++ b/src/components/tree-select.tsx
@@ -2,17 +2,20 @@ import { ReactNode, useCallback, useState } from 'react';
 import cn from 'classnames';
 import { Except } from 'type-fest';
 
-import {
-  BasicItem,
-  getFlattenedPaths,
-  restructureFlattenedTreeDataForAutocomplete,
-} from '../utils';
 import DropdownButton, { DropdownButtonProps } from './dropdown-button';
 import Button from './button';
 import Autocomplete from './autocomplete';
 
-import '../styles/components/tree-select.scss';
+import {
+  BasicItem,
+  getFlattenedPaths,
+  restructureFlattenedTreeDataForAutocomplete,
+  getSingleChildren,
+} from '../utils';
+
 import { AutocompleteItemType } from './autocomplete-item';
+
+import '../styles/components/tree-select.scss';
 
 export type TreeSelectProps<Item extends BasicItem<Item>> = {
   /**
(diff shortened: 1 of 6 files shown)


src/components/tree-select.tsx
```diff
@@ -56,19 +59,23 @@ const TreeSelect = <Item extends BasicItem<Item>>({
...
-    },
-    [openNodes]
+    (node: AutocompleteItemType | BasicItem<Item>) =>
+      setOpenNodes((openNodes) => {
+        const index = openNodes.indexOf(node.id);
+        if (index === -1) {
+          return [...openNodes, node.id];
+        }
+        return openNodes.filter((id) => id !== node.id);
+      }),
```
dlrice, line 77:
```suggestion
      setOpenNodes((openNodes) =>
        openNodes.includes(node.id)
          ? openNodes.filter((id) => id !== node.id)
          : [...openNodes, node.id]
      ),
```
Just suggesting because I initially thought something is done with `index` other than just checking inclusion

src/components/tree-select.tsx
```diff
@@ -56,19 +59,23 @@ const TreeSelect = <Item extends BasicItem<Item>>({
...
-      }
-    },
-    [openNodes]
+    (node: AutocompleteItemType | BasicItem<Item>) =>
+      setOpenNodes((openNodes) => {
+        const index = openNodes.indexOf(node.id);
+        if (index === -1) {
+          return [...openNodes, node.id];
+        }
+        return openNodes.filter((id) => id !== node.id);
```
dlrice, line 76:
```suggestion
        return [...openNodes.slice(0, index), ...openNodes.slice(index + 1)];
```
(related to comment below) Or could use the index since we already have it.

dlrice (approved):
This is a nice improvement - the animation makes it look really sharp.

1. The dropdown button component arrow looks different to the tree select node arrows - could be nice to copy over what you've done here for the dropdown button. Not necessarily with the animation because the dropdown button starts down while the tree select nodes start sideways. Unless we want the tree select nodes to start down then transition to sideways?
2. In my testing came across double scrollbars (which existed before this PR):
Screenshot 2021-08-18 at 19 12 11
which I believe comes from
```
  .dropdown-menu__panel {
    overflow-y: auto;
    max-height: 70vh;
  }
```
not sure if there is a nice way to resolve this? I can create a jira.

src/components/tree-select.tsx
aurel-l: Something was being done with the index, because I was planning on replacing the previous `slice` by a `splice`, but then I switched to filter because I though it was easier to understand the meaning behind what was being done and forgot to remove that index.
I'll take your suggested change.

src/components/tree-select.tsx
aurel-l: Went with the previous suggestion instead of this one

aurel-l:
>     1. The dropdown button component arrow looks different to the tree select node arrows - could be nice to copy over what you've done here for the dropdown button. Not necessarily with the animation because the dropdown button starts down while the tree select nodes start sideways. Unless we want the tree select nodes to start down then transition to sideways?

Not sure I'd touch that now, because it's in another component altogether, and the triangle is coming from a mixin possibly reused in other places that I don't really want to touch right now. But I'm all for using unicode symbols when they exist like I did here instead of bending CSS borders in order to make triangles.

>     2. In my testing came across double scrollbars (which existed before this PR):
...
> not sure if there is a nice way to resolve this? I can create a jira.

Yep, that was already covered in this Jira https://www.ebi.ac.uk/panda/jira/browse/TRM-26452 but since you found the source of the problem I removed the line of code responsible for that in a commit and we'll be tackling 3 jiras instead of 2 then!

dlrice:
> Not sure I'd touch that now, because it's in another component altogether, and the triangle is coming from a mixin possibly reused in other places that I don't really want to touch right now. But I'm all for using unicode symbols when they exist like I did here instead of bending CSS borders in order to make triangles.

Yeah, could be something to discuss at a design meeting.

> Yep, that was already covered in this Jira https://www.ebi.ac.uk/panda/jira/browse/TRM-26452 but since you found the source of the problem I removed the line of code responsible for that in a commit and we'll be tackling 3 jiras instead of 2 then!

Nice that the fix was that easy!

PR #225, opened 2021-09-03
Info list item link

dlrice:
## Purpose
* Allow info list item to be a link.
* Don't show border when view is small and items become rows.
* Use content for border as presumably this will be taller.

**BEFORE**
Screenshot 2021-09-03 at 13 21 29

**AFTER**
Screenshot 2021-09-03 at 13 21 35

## Approach
Copy/pasted approach used in Card

## Testing
Added unit test

## Stories
Added info list item with a link.

## Checklist

### DIFF
--- a/src/styles/components/decorated-list-item.scss
+++ b/src/styles/components/decorated-list-item.scss
@@ -98,9 +98,10 @@ $alt-decorative-bar: 0.25rem solid $colour-pastel-blue;
     cursor: pointer;
   }
 
+  // Lifted from card style
   &--has-link {
     transition: 0.5s background-color ease;
-    position: relative;
+    position: relative; // Needed this otherwise the whole list becomes the anchor element
 
     &:hover,
     &:focus-within {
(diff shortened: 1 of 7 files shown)


src/styles/components/decorated-list-item.scss
```diff
@@ -97,6 +131,7 @@ $alt-decorative-bar: 0.25rem solid $colour-pastel-blue;
     &__content {
       border-left: none;
       padding-left: 0;
+      padding-left: 0;
```
aurel-l, line 134:
```suggestion
      padding-left: 0;
```

Might be unwanted I believe

aurel-l (approved):
Good to go 👍🏽 
We need to keep in mind that this new link is not accessible, so we will still need to have an actual accessible link within this component (in the title I guess) corresponding to the same thing this hidden link is pointing too. That needs to be done in the uniprot-website logic I guess.
That was the use case in the card where we had a link in the title and the whole card was *also* clickable.
It means that when we tab through the page we'll have to be able to find the link within because we won't be able to find the one that is hidden in this component.

PR #236, opened 2022-01-19
sliding panel, reason for closing, and closing through escape key

aurel-l:
## Purpose & approach
To cater for the needs of the contextual help, pass the reason for which the `onClose` callback has been called
Also, add support for closing through the escape key (see end of PR description)

## Testing
Added new tests

## Stories
sliding-panel

## Checklist

~Pending question:~
~To enable the "press Escape key" behaviour, I need to listen to the key down event. It is only emitted from within focusable elements. There are multiple ways to handle this:~
 - ~Make the sliding panel focusable (but it will appear in the tabbing flow), event listener on the sliding panel. Any key press inside will be picked up -> current approach in this draft~
 - ~Event listener on the sliding panel, relying on the focus being moved to the first focusable element within automatically, event bubbling will make it so that the sliding panel gets the key press (but what if there are no focusable element within? Not the case afaik)~
 - Event listener on the whole document, any escape key pressed when focus inside or outside the sliding panel will make it close

~And maybe another approach could be possible.~

### DIFF
--- a/src/components/__tests__/sliding-panel.spec.tsx
+++ b/src/components/__tests__/sliding-panel.spec.tsx
@@ -10,7 +10,7 @@ describe('SlidingPanel component', () => {
     const onClose = jest.fn();
     renderWithRouter(
       <>
-        <div data-testid="outside-component" />
+        <div data-testid="outside-element" />
         <SlidingPanel onClose={onClose} position="left" title="Title">
           Sliding panel content
         </SlidingPanel>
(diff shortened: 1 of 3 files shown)


xwatkins:
> Event listener on the whole document, any escape key pressed when focus inside or outside the sliding panel will make it close

I would expect this to be the behaviour. I might want to have a panel that doesn't close on outside click but closes on escape even if the focus is lost. What do you think?

xwatkins (approved):

PR #238, opened 2022-01-25
Add tippy, tooltips to column headers

xwatkins:
## Purpose
Add tooltips functionality to column headers.

## Approach
Added dependency on Tippy (and react Tippy). Had to add a `forwardRef` to make it work with the column headers.

## Testing
Update snapshots.

## Stories
Updated default content to add 2 examples (first 2 columns)

## Checklist

### DIFF
--- a/src/components/sequence.tsx
+++ b/src/components/sequence.tsx
@@ -20,6 +20,8 @@ import {
   Button,
 } from '.';
 
+import { InfoListItem } from './info-list';
+
 import aminoAcidsProps from './data/amino-acid-properties.json';
 
 import '../styles/components/sequence.scss';
@@ -128,10 +130,7 @@ type SequenceProps = {
   /**
    * Data to be displayed in an InfoData component above the sequence
    */
-  infoData?: {
-    title: ReactNode;
-    content: ReactNode;
-  }[];
+  infoData?: InfoListItem[];
   /**
    * The URL to download the isoform sequence
    */
--- a/src/components/sequence.tsx
+++ b/src/components/sequence.tsx
@@ -129,7 +129,7 @@ type SequenceProps = {
    * Data to be displayed in an InfoData component above the sequence
    */
   infoData?: {
-    title: string;
+    title: ReactNode;
     content: ReactNode;
   }[];
   /**
--- a/src/styles/components/data-table.scss
+++ b/src/styles/components/data-table.scss
@@ -173,4 +173,10 @@
       white-space: nowrap;
     }
   }
+
+  // Tippy link styling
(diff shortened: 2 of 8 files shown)


dlrice (commented):
With the current tooltip styling our links don't have much contrast on the dark background:

Screenshot 2022-01-25 at 18 07 06

I've had a play around with `theme="light-border"`:

Screenshot 2022-01-25 at 18 04 21

And that looks OK? Or maybe we just need to have a light style applied to links inside of tooltips?

Note that themes needs eg `import 'tippy.js/themes/light-border.css';`

xwatkins:
> With the current tooltip styling our links don't have much contrast on the dark background:
> 
> Screenshot 2022-01-25 at 18 07 06
> 
> I've had a play around with `theme="light-border"`:
> 
> Screenshot 2022-01-25 at 18 04 21
> 
> And that looks OK? Or maybe we just need to have a light style applied to links inside of tooltips?
> 
> Note that themes needs eg `import 'tippy.js/themes/light-border.css';`

I would argue that the dark background has more contrast and is more easily identifiable than the light one?

xwatkins:
Maybe we could go with the default and discuss with Marija when she's back.

dlrice:
> Maybe we could go with the default and discuss with Marija when she's back.

Yeah, I agree we will need Marija's input on this. Just to make it easier for me to quickly see the links as I add the tooltip content could we go with a light theme for now? It takes a bit of effort to see that a link is present and/or what the link text is.

xwatkins:
Ah sorry I had missed the link altogether! I will restyle the link to be white with underline :-)

src/components/sequence.tsx
```diff
@@ -129,7 +129,7 @@ type SequenceProps = {
    * Data to be displayed in an InfoData component above the sequence
    */
   infoData?: {
-    title: string;
+    title: ReactNode;
     content: ReactNode;
   }[];
```
aurel-l, line 134: If this is to be passed to an InfoList component. It might be better to take the type from there to avoid inconsistencies later on.

src/styles/components/data-table.scss
```diff
@@ -173,4 +173,10 @@
       white-space: nowrap;
     }
   }
+
+  // Tippy link styling
+  .tippy-content a.external-link {
```
dlrice, line 178:
```suggestion
  .tippy-content a {
```
Column tooltips will also have internal links to `/help`

xwatkins: `a.external-link` is more specific so `a` alone wouldn't override it. Added both.

PR #239, opened 2022-03-18
2022 04 dropdown header refactor

aurel-l:
## Purpose
Refactor Header and Dropdown components

## Approach
- Refactor dropdown, create a controlled and uncontrolled version (see similar approach in tabs)
- Replace header to take a slotting approach instead of configuration props that are not really flexible. Pass elements as props
  - Allows to use a link controlled from the outside of Franklin -> can pass the referrer to the contact form
- Create new Dropdown components:
  - ControlledDropdown: external control of open/close state
  - Dropdown: automatic logic (close on click outside), specify a prop to close (outside state triggering a closing), use ControlledDropdown internally
  - Fixes alignment issue, especially with the triangle marker. Make it change depending on state
  - Rely on a slotting approach too, similar to the header changes
  - Keep previous DropdownButton because changes were too complex to in the last place it's used: TreeSelect.
- Misc: fix facet close button, prevent overlapping with facet text
- Misc: MainSearch, on regular and smaller screens, make the secondary content (used for the "Advanced | List" buttons) disappear when the text input is focused. Still tab-accessible. Gives more space to the text search when interacting with it. 

## Testing
Updated snapshot, manual testing

## Stories
- Dropdown
- Header

## Checklist
  - Lower test coverage, hopefully mainly because I removed logic


aurel-l:
Not really happy to not get it through proper code review, but it was unfinished anyway (touching TreeSelect was too complicated). But it contains visual fixes that were around and reported since a while ago, and allows improvements that will be implemented in https://github.com/ebi-uniprot/uniprot-website/pull/578
Still creating it as a PR just to group everything together whenever we come back to it.
Plenty of issues regarding specificity fights, starting even from too high specificity inside of Foundation itself. Might need to reconsider some of it whenever we decide to go forward with removing Foundation altogether.

PR #240, opened 2022-03-30
Tree select with tags

dlrice:
## Purpose
In the tree select for field names there are some fields which need to be found by text which is not in the field name. To deal with this an additional attribute `tags` has been created which is an array of text which can be matched against. This PR accommodates the `tags` attribute.

## Approach

- Include tags when finding matching nodes
- Renamed a few things to make it clearer what they do
- CI tests failing because of not having merging against the latest target branch so fixed this

## Testing

- Added unit tests
- Linked to uniprot-website's main branch and works

## Stories
Not updated

## Checklist


aurel-l (approved):
I did a small modification in a test in order to make it clear because it found the searched value in the tags and not in the label

PR #254, opened 2023-03-02
Substring highlight

dlrice:
## Purpose
Update highlight style as per discussion with Minjoon.

## Approach
Refactored as a functional component and defined new style in a scss file.

## Testing
Updated

## Stories
N/A

## Checklist


src/components/autocomplete-item.tsx
```diff
@@ -40,9 +40,14 @@ const AutocompleteItem = ({
...
         className={active ? 'hover' : ''}
       >
-        {substringToHighlight
-          ? highlightSubstring(item.pathLabel, substringToHighlight)
-          : item.itemLabel}
+        {substringToHighlight ? (
+          <SubstringHighlight
+            string={item.pathLabel}
+            substring={substringToHighlight}
+          />
```
aurel-l, line 47: I'd prefer a component API similar to other components where we modify text before rendering it (e.g. LongNumber) where the text is a child of the component.
```suggestion
          <SubstringHighlight substring={substringToHighlight}>
            {item.pathLabel}
          </SubstringHighlight>
```
We can enforce the type of children to be only a string in this case

dlrice: Good point

src/styles/components/substring-highlight.scss
```diff
@@ -0,0 +1,11 @@
+@import '../colours';
+
+.highlight {
+  font-weight: bold;
+  background-color: rgba(
+    red($colour-medium-turquoise),
+    green($colour-medium-turquoise),
+    blue($colour-medium-turquoise),
+    0.4
+  );
```
aurel-l, line 10: I think I've used something like
```suggestion
  background-color: rgba(
    $colour-medium-turquoise,
    0.4
  );
```
in the past and I believe it should work too

src/styles/components/substring-highlight.scss
```diff
@@ -0,0 +1,9 @@
+@import '../colours';
+
+.highlight {
```
aurel-l, line 3: Do we want that to be part of the design system in general?
```suggestion
.highlight, mark {
```
Would allow us to use `<mark>` anywhere we want (not necessarily with the substring component) and have consistent styling

dlrice: Also might make more sense to move that style into a `styles/common/_mark.scss`

aurel-l: We could, although I'm afraid we might lose the link between the component and the style in a refactor, or not load the correct style whenever needed.
In that case I'd put this style as part of the global franklin styles and not in a file that is loaded by this component

aurel-l: Right, we crossed our replies, but `styles/common/_mark.scss` approach looks good to me

dlrice: Updated

aurel-l (approved):

PR #263, opened 2024-03-28
Upgrade react et al.

dlrice:
## Note

There are merge conflicts arising in gitlabci. I can't reproduce locally so I think let's just merge when it's reviewed and resolve at that point.

## Changes

- Uses Node v20
- Still uses yarn
- Upgrade react to v18
	- Type change: FC now needs to be FC<PropsWithChildren>
- Upgrade storybook. The story format has changed eg `type Story = StoryObj<typeof Accordion>`
- Update react testing library
- Removed DisplayMenu as unused
- Remove react router and leave this for the consumer (uniprot website).
	- Accepts `<link />`:
		- Facets
		- Card
		- Decorated List Item
		- Tile
		- Header
	- Moved to uniprot:
		- Facet
		- In page nav
	- Needs pathname:
		- Sliding Panel
- useDataCheckboxes due to type issues "A callback ref should not return a function" it now stores checkbox references in state and watches for changes with useEffect
- Data list/table return JSX.Element
- `/* eslint-disable react/no-unused-prop-types */` for data-list and data-table because of [eslint false positive](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unused-prop-types.md#false-positives-sfc)
- Update node version for gitlabci pipeline
- Finish updating Stories for new format
- Use new `useDeferredValue` (lets you defer updating a part of the UI) in `AccordionSearch`


src/components/accordion-search.tsx
```diff
@@ -222,51 +214,17 @@ const AccordionSearch = ({
...
-        { leading: true }
-      ),
-    [accordionData]
-  );
-
-  useEffect(
-    () => debouncedFilterAccordionData.cancel,
-    [debouncedFilterAccordionData]
-  );
+  const deferredFilteredAccordionData = useDeferredValue(filteredAccordionData);
```
aurel-l, line 220: Have you performance-checked with a big dataset to see its effect?
I'm curious to see how much we gain from using this.

I'm also wondering if the slowdown we were sometimes viewing before was due to the re-rendering (I think so), or to the `filterAccordionData` call lower in the event handler, which then might not be improved by the useDeferred.

Asking these things more out of curiosity than anything as I haven't had the chance to play with these yet

stories/Histogram.stories.tsx
```diff
@@ -32,17 +34,95 @@ const gaussianMax = Math.max(...gaussianSample);
...
+      binSize: number;
+    }
+>;
+
+const meta: Meta<WithNBins> = {
+  component: HistogramComponent,
+  title: 'Visualisation/Histogram',
+  parameters: {
+    purposeFunction: {
+      purpose: 'purpose',
```
Swaathik, line 60: I noticed they were empty in few places. Shall we remove it or leave it empty?

package.json
```diff
@@ -28,9 +28,10 @@
     "prettier": "prettier src --write",
     "prerelease": "yarn test && yarn build",
     "release": "yarn publish && git push --follow-tags origin main",
-    "start": "start-storybook -p 6006",
-    "build-storybook": "build-storybook -c .storybook -o .out",
-    "prepare": "husky install"
+    "start": "storybook dev -p 6006",
+    "build-storybook": "storybook build",
+    "prepare": "husky install",
+    "storybook": "storybook dev -p 6006"
```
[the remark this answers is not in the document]
dlrice: Removed

stories/Histogram.stories.tsx
dlrice: Removed everywhere

src/components/accordion-search.tsx
dlrice: I'll link to the website and throttle my cpu to see how it looks. I'm thinking that with the `useDeferred` even if it is `filterAccordionData` that takes a lot of time shouldn't the previous `filteredAccordionData` still be present on the screen and so help keep a smooth(er) UI?

aurel-l (approved):
Approving the code changes here, pending full review on the uniprot-website side of things, and also after checking the gitlab issue which appears to just be git conflicts (although quite a lot of them) but this is still strange because there's not been new work on `main` since you branched out of it... 🤔

aurel-l (approved):

PR #266, opened 2025-02-11
Tab refactor

dlrice:
## Purpose
[Restyle tabs and a few bug fixes](https://www.ebi.ac.uk/panda/jira/browse/TRM-32429)

## Approach
- Create `bordered` prop to have the tab header bordered. This is false by default.
- Create `disabled` prop.
- Use flexbox and small adjustments to style

## Testing
None

## Stories
Added more

## Checklist

### DIFF
--- a/src/components/tabs.tsx
+++ b/src/components/tabs.tsx
@@ -159,11 +159,11 @@ export const Tabs = ({
           ({
             title,
             id,
-            className, // eslint-disable-line no-shadow
+            className,
             children: _,
             defaultSelected: __,
             cache: ___,
-            ...props // eslint-disable-line no-shadow
+            ...props
           }) =>
             title && (
               <div
--- a/src/components/tabs.tsx
+++ b/src/components/tabs.tsx
@@ -33,6 +33,9 @@ type TabProps = {
    * Option to render and hide tab (display:none) rather than remove from the DOM
    */
   cache?: boolean;
+  /**
+   * Option to disable selection of tab
+   */
   disabled?: boolean;
 } & Except<HTMLAttributes<HTMLDivElement>, 'title' | 'id'>;
 
@@ -54,6 +57,9 @@ type TabsProps = {
    * <Tab>
    */
   active?: string | number;
+  /**
+   * Optional bordered styling of tab headers
+   */
   bordered?: boolean;
 } & Except<HTMLAttributes<HTMLDivElement>, 'children'>;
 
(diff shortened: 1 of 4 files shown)


src/styles/components/tabs.scss
```diff
@@ -4,28 +4,65 @@
 .tabs {
   &__header {
     width: 100%;
-    white-space: nowrap;
+    display: flex;
+    flex-direction: row;
+    flex-wrap: nowrap;
     overflow-x: auto;
-    border-bottom: 0.0625rem solid $colour-platinum;
+    box-shadow: inset 0 -0.1rem 0 0 $colour-platinum;
```
aurel-l, line 11: It might be good, while touching this file, to replace all SASS colour variables with the franklin Custom Properties that we have defined.

src/styles/components/tabs.scss
```diff
@@ -4,28 +4,65 @@
...
+        &:first-child {
+          border-left: 0.1rem solid $colour-platinum;
+        }
+      }
+
+      &--active,
       &:hover,
       &:focus {
-        color: $colour-sea-blue;
+        background-color: rgba($colour-pastel-blue, 0.19);
```
aurel-l, line 39: And in this case, if you want to explore new CSS features, have a look at this page https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color#relative_value_syntax
I think this would end up being `background-color: color(from var(--fr--color-pastel-blue) srgb r g b / 0.19);`.

Support is fairly good now https://caniuse.com/css-color-function but if you want you can provide a simpler fallback colour in the line before in case it's not supported

src/components/tabs.tsx
```diff
@@ -168,12 +178,17 @@ export const Tabs = ({ children, active, className, ...props }: TabsProps) => {
...
                   'tabs__header__item',
                   {
+                    'tabs__header__item--bordered': bordered,
+                    'tabs__header__item--disabled': disabled,
                     'tabs__header__item--active': id === activeFromPropsOrState,
                   },
                   className
                 )}
+                aria-disabled={disabled ? true : undefined}
+                onFocus={disabled ? (event) => event.target.blur() : undefined}
```
aurel-l, line 188: I'm not sure that this is needed, because of the tabIndex -1 below

aurel-l: I tried locally with and without this line and didn't see any difference, but maybe it was done for a reason I'm missing?

aurel-l (approved):

aurel-l:
Unrelated to this PR, but I've recently seen https://www.typescriptlang.org/tsconfig/#declarationMap that might be useful to add, so that tab clicking on Franklin component would point to the code instead of the declaration file. We could try it with the next Franklin release and see if it improves DX in uniprot-website

